Issue #7, opened 2019-06-05
Open tracing Tags


tbrown1979:
In attemping to add an http4s module I found that adding the tags from OpenTracing could be useful for defining the tag names for http method, status code, etc...

These are defined here: https://github.com/opentracing/specification/blob/master/semantic_conventions.md

I think it would make sense for these to go in `core`, but let me know. I'm more than willing to get this added to the project 😄

tpolecat:
That makes sense. You can put them in core for now. Note that I just rewrote basically the whole library, so be sure you're up to date.

Also note that they just that OpenTracing + OpenCensus are merging and becoming OpenTelemetry, so I imagine there will be a lot of changes. I'm trying to ignore all that and just work with Honeycomb for now.

PR #8, opened 2019-06-06
Add OpenTracing Tags

tbrown1979:
This PR is addressing #7 

This isn't really final yet, but I wanted to get it up for review. Let me know if you'd like things structured differently 😄 I added a `Field` case class to represent the `(String, TraceValue)` tuple, but before I added it elsewhere I figured I'd post this and see what you thought. 

I'm wondering if a `Trace` and `Span` should make a distinction between logging and tagging. Right now it's just `put` which, in the Jaeger version, just does `.setTag`. They do ultimately follow the same `key:value` format though. Let me know what you think!

### DIFF
--- a/modules/core/src/main/scala/Tags.scala
+++ b/modules/core/src/main/scala/Tags.scala
@@ -7,88 +7,88 @@ package natchez
 object Tags {
 
   // The software package, framework, library, or module that generated the associated Span. E.g., "grpc", "django", "JDBI".
-  def Component(c: String) = Field("component", c)
+  def component(c: String) = ("component", c)
 
-  object Db {
+  object db {
     private val prefix = "db"
 
     // Database instance name.
     // E.g., In java, if the jdbc.url="jdbc:mysql://127.0.0.1:3306/customers", the instance name is "customers".
-    def Instance(i: String) = Field(s"$prefix.instance", i)
+    def instance(i: String) = (s"$prefix.instance", i)
 
     // A database statement for the given database type.
     // E.g., for db.type="sql", "SELECT * FROM wuser_table"; for db.type="redis", "SET mykey 'WuValue'".
-    def Statement = Field(s"$prefix.statement", _)
+    def statement(s: String) = (s"$prefix.statement", s)
 
     // Database type.
     // For any SQL database, "sql". For others, the lower-case database category, e.g. "cassandra", "hbase", or "redis".
-    def Type(t: String) = Field(s"$prefix.type", t)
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/Tags.scala
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2019 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package natchez
+
+object Tags {
+
+  // The software package, framework, library, or module that generated the associated Span. E.g., "grpc", "django", "JDBI".
+  def Component(c: String) = Field("component", c)
```
tpolecat: What if instead of `Field` these just returned a `(String, TraceValue)` pair? That way you could mix and match known tags and arbitrary pairs, like

```scala
Trace[F].put(Component("foo"), "bar" -> 42, ...)
```

modules/core/src/main/scala/Tags.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  object Db {
+    private val prefix = "db"
+
+    // Database instance name.
+    // E.g., In java, if the jdbc.url="jdbc:mysql://127.0.0.1:3306/customers", the instance name is "customers".
```
tpolecat: Can we make these ScalaDoc comments? They seem useful for users.

modules/core/src/main/scala/Tags.scala
```diff
@@ -0,0 +1,94 @@
...
+    def Instance(i: String) = Field(s"$prefix.instance", i)
+
+    // A database statement for the given database type.
+    // E.g., for db.type="sql", "SELECT * FROM wuser_table"; for db.type="redis", "SET mykey 'WuValue'".
+    def Statement = Field(s"$prefix.statement", _)
```
tpolecat: The uppercase method names weird me out. I think I might prefer the inner objects and method names match up exactly with the tags they create, which means lowercasing the identifiers and adding underscores in a few places; i.e., `http.status_code(200)` would yield `"http.status_code" -> 200`. What do you think?

tpolecat: Can we take a string explicitly here?

tpolecat:
Looks good. Can you turn those comments in to scaladoc style?

tbrown1979:
@tpolecat Yep, I can hopefully get that done tonight! Just addressing things as I get a bit of time here and there 😄

tbrown1979:
@tpolecat Comments should be all addressed 😃

PR #11, opened 2019-06-08
add continueOrElseRoot, cross-compile for 2.11

tpolecat:
This adds a `continueOrElseRoot` method that tries to continue a trace, and if the headers aren't present it will fall back to creating a new root span. Also adds 2.11 support, which I will be happy to ditch at the slightest provocation but it works fine for now.


tbrown1979:
Nice addition! :)

PR #12, opened 2019-07-07
Add support for OpenCensus traces

janstenpickle:
Loving the project! I just wanted to add support [OpenCensus](https://opencensus.io). I realise there is a [OpenTelemetry](https://opentelemetry.io) coming along which should subsume OpenCensus and OpenTracing, however it's not clear how long it will be until it's usable.

OpenCensus supports pluggable exporters, but I've only added an implementation for the standard agent. Unfortunately it seems the way exporters are configured and resolved is a bit side effecty and uses singletons, so we can't stop someone configuring new exporters outside of a resource.

I'm not sure what formatting standard you were using, so I just used the default scalafmt config. I can add a different config if you'd like?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ This is an minimal distributed tracing effect for Cats, inspired by earlier work
 
 - [Jaeger](https://www.jaegertracing.io/)
 - [Honeycomb](https://www.honeycomb.io/)
+- [OpenCensus](https://www.opencensus.io/)
 
 It supports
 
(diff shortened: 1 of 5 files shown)


tpolecat:
Thanks! This looks good. Can you add doc comments that explain the edge case re: singletons?

I'm using the "tpolecat likes it" formatter which has no machine-readable representation. What you have here is fine.

janstenpickle:
> This looks good. Can you add a doc comment that explain the edge case re: singletons?

Thanks, will do!

> I'm using the "tpolecat likes it" formatter which has no machine-readable representation.
 
😆

PR #13, opened 2019-07-28
Add support for Lightstep

mijicd:
### Summary

- Introduced lighstep bindings for both gRPC and HTTP.
- Fixed the configurable parameters to the ones enlisted [here](https://github.com/lightstep/lightstep-tracer-java/tree/master/lightstep-tracer-jre-bundle#parameters).

### Remarks

- Configuration can be made more specific via [refined](https://github.com/fthomas/refined).
- `JaegerSpan` and `LightstepSpan` are the same, while entrypoints are very similar. Should they be kept as they are or extracted somewhere?
- `oraclejdk8` has been replaced with `openjdk8` due to TravisCI failures as described in [this thread](https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766)

### DIFF
--- a/modules/lightstep/src/main/scala/Lightstep.scala
+++ b/modules/lightstep/src/main/scala/Lightstep.scala
@@ -5,13 +5,16 @@
 package natchez
 package lightstep
 
-import cats.effect.{ Sync, Resource }
+import cats.effect.{ Resource, Sync }
+import cats.syntax.applicative._
 import com.lightstep.tracer.jre.JRETracer
 import com.lightstep.tracer.shared.Options
 import io.opentracing.Tracer
+import io.opentracing.propagation.{ Format, TextMapAdapter }
+
+import scala.collection.JavaConverters._
 
 object Lightstep {
-  // TODO: use refined types to properly specify the config
   final case class Configuration(
     accessToken: String,
     componentName: String,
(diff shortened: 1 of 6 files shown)


modules/lightstep/src/main/scala/LightstepSpan.scala
```diff
@@ -0,0 +1,40 @@
...
+package lightstep
+
+import cats.effect.{ Resource, Sync }
+import cats.implicits._
+import io.{ opentracing => ot }
+import io.opentracing.propagation.{ Format, TextMapAdapter }
+
+import scala.collection.JavaConverters._
+
+private[lightstep] final case class LightstepSpan[F[_]: Sync](
```
jcranky: What here is lightstep specific? This whole file is `opentracing` related, but vendor neutral, isn't it?

mijicd: Yes, that's why I mentioned it in the PR description. Having in mind the current library API, I'd probably extract them outside of vendor-specific implementations.

jcranky: Ah, I somehow missed the `JaegerSpan` comment, thanks for pointing it out. I would definitely vote for having it extracted to a single place, to also make it clear that it could be used for more OpenTracing vendors in the future.

tpolecat:
Thanks for the contribution, I will try to have a look later today.

modules/lightstep/src/main/scala/LightstepSpan.scala
tpolecat: The problem here is that Jaeger and LightStep use different, incompatible versions of OpenTracing; so we would need at least two OpenTracing layers here. Maybe they'll get their act together at some point but for now I think it's ok to bury OpenTracing and just do vendor modules.

This is all out the window when OpenTelemetry shows up.

modules/lightstep/src/main/scala/Lightstep.scala
```diff
@@ -0,0 +1,62 @@
...
+object Lightstep {
+  final case class Configuration(
+    accessToken: String,
+    componentName: String,
+    collectorHost: String,
+    collectorProtocol: String,
+    collectorPort: Int
+  )
+
+  def entryPoint[F[_]: Sync](config: Configuration): Resource[F, EntryPoint[F]] =
```
tpolecat: I think maybe we should take an `OptionsBuilder => F[Tracer]` here, like we do with Jaeger, so we're not in the business of configuring LightStep ourselves. Does that makes sense?

mijicd: Actually, it does, as it is even more flexible. However, I would accept only the `OptionsBuilder` and change the Jaeger to accept the `Configuration` as service name can still be overridden which kinda defeats its purpose.

mijicd: Once I think about it, I'll go with `OptionsBuilder => F[Tracer]`, but without setting any parameters during creation.

tpolecat:
👍

PR #18, opened 2019-08-30
WIP: propagation scopes

### LINKED ISSUE #17: Propagate span values to children spans
There should be possible to automatically propagate some values (e.g. correlation ids) from parent to its children. I propose to make this the default behavior of `put`, and to add `putLocal` which should be used to add span-local values.

tpolecat:
This updates `put` to take a `Propagation` value that explains how the fields should be propagated to child spans. Resolves #17 

Value | Meaning
-|-
`Propagation.None`|  fields will not be propagated to child spans
`Propagation.Local`| fields will be propagated to child spans, but will not cross network boundaries
`Propagation.Extended` | fields will be propagated to child spans, and across network boundaries

Partially implemented for Honeycomb. Parameter currently ignored for other back ends.

### DIFF
--- a/modules/core/src/main/scala/Span.scala
+++ b/modules/core/src/main/scala/Span.scala
@@ -10,7 +10,7 @@ import cats.effect.Resource
 trait Span[F[_]] {
 
   /** Put a sequence of fields into this span. */
-  def put(fields: (String, TraceValue)*): F[Unit]
+  def put(propagation: Propagation, fields: (String, TraceValue)*): F[Unit]
 
   /**
    * The kernel for this span, which can be sent as headers to remote systems, which can then
@@ -20,5 +20,20 @@ trait Span[F[_]] {
 
   /** Resource that yields a child span with the given name. */
   def span(name: String): Resource[F, Span[F]]
+
+}
+
+
+sealed trait Propagation extends Product with Serializable
+object Propagation {
+
+  /** Fields will not be propagated to child spans. */
+  case object None     extends Propagation
+
+  /** Fields will be propagated to child spans, but will not cross network boundaries. */
+  case object Local    extends Propagation
+
+  /** Fields will be propagated to child spans, and across network boundaries. */
+  case object Extended extends Propagation
 
 }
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/Span.scala
```diff
@@ -22,3 +22,18 @@ trait Span[F[_]] {
...
 }
+
+
+sealed trait Propagation extends Product with Serializable
+object Propagation {
+
+  /** Fields will not be propagated to child spans. */
+  case object None     extends Propagation
+
+  /** Fields will be propagated to child spans, but will not cross network boundaries. */
```
DougC, line 33: Is there a case for propagating Local fields back out to parent spans as well as into child spans? For example, I have a situation where I'm inserting an entity into the db, allocating a primary key at that point, and I'm adding the generated key to the span that wraps the db call. But it would be valuable if that key also appeared on the parent spans so that I can easily cross-reference them.

I can see, though, that there would be cases to address such as what to do if the parent span already has a value for a Local field set by a child span.

modules/core/src/main/scala/Span.scala
```diff
@@ -22,3 +22,18 @@ trait Span[F[_]] {
...
+  case object Local    extends Propagation
+
+  /** Fields will be propagated to child spans, and across network boundaries. */
+  case object Extended extends Propagation
```
DougC, line 37: Do you have any thoughts as to how these Extended params would be made available to things like the http4s middleware so that they can be propagated across the network boundary? I imagine one way is that they could be merged onto the Kernel that the middleware pulls off the trace.

Also, any thoughts about the mechanism the http4s middleware might use to do the propagation? Each field could become a header in the request (perhaps prefixing with `X-Natchez-Field-` to avoid fields being set with names that clash with http headers?) or a single `X-Natches-Fields` header that encodes the name/value pairs into a single string?

One specific thing I'd really like to see, though, is a way to add a field that ends up being set as the `X-Correlation-Id` header on http calls. Maybe that ends up as logic in the middleware to extract a value with a known name from the kernel.

tpolecat: Yes, they become part of the kernel. Prefixing with `X-Natchez-Field` is a good idea. We could set the trace ID as the correlation ID although this would need to be optional because it's an existing header people may already be using. I don't really see the point of using both tracing and a correlation ID but I guess people do it.

PR #19, opened 2019-09-01
Add instance of Trace[Resource[F, ?]]

iRevive:
Currently, it's impossible to use Trace with Resource (until explicit Trace[Resource[F, ?]] is imported). 
From my point of view, it can be a part of the library. 

Real-world use case:
```scala
def create[F[_]: Concurrent: Timer: ContextShift: Trace](config: PostgresConfig): Resource[F, HikariTransactor] = 
  Trace[Resource[F, ?]].span("create-transactor")(createTransactor(config))

private def createTransactor[F[_]: Concurrent: Timer: ContextShift](config: PostgresConfig): Resource[F, HikariTransactor[F]] =
  for {
    ce <- ExecutionContexts.fixedThreadPool[F](32)
    te <- ExecutionContexts.cachedThreadPool[F]
    xa <- HikariTransactor.newHikariTransactor[F](config.driver, config.uri, config.user, config.password, ce, te)
  } yield xa
```


modules/core/src/main/scala/Trace.scala
```diff
@@ -86,4 +86,18 @@ object Trace {
...
+    new Trace[Resource[F, ?]] {
+
+      override def put(fields: (String, TraceValue)*): Resource[F, Unit] =
+        Resource.liftF(T.put(fields: _*))
+
+      override def kernel: Resource[F, Kernel] =
+        Resource.liftF(T.kernel)
+
+      override def span[A](name: String)(fa: Resource[F, A]): Resource[F, A] =
+        Resource(T.span(name)(fa.allocated))
```
tpolecat: Hm, it seems like we want the extent of the span to be `use` rather than `allocated`.

iRevive: I'm not sure that `fa.use(...)` can be used because there is no available `A => F[B]`  function for `use` method. 

I've added some tests. Looks like this implementation works correctly.

kubukoz:
As far as I can see, this makes the span live as long as the resource is being allocated. I can imagine some cases when people will want to have a span for the whole lifetime of the resource, or for its cleanup. How about this - we would wrap a resource and create spans for:

- lifetime, with children:
- - allocating
- - using
- - closing

iRevive:
@kubukoz thanks for the feedback. 

The current implementation creates a span exclusively for the allocation function. 
I can create separate spans for the `acquire` and `release` functions:
```scala
override def span[A](name: String)(fa: Resource[F, A]): Resource[F, A] = {
  val resource = T.span(name + "-acquire")(fa.allocated).map {
    case (acquire, release) =>
      (acquire, T.span(name + "-release")(release))
  }

  Resource(resource)
}
```

But I cannot create a span for the `use` function since `Resource` does not provide this kind of interception, or am I missing something?

kubukoz:
Yeah, I think that's about right, sorry 😅 we can't really trace the `use` part, but I think we can create a parent span for the whole resource if we create a common parent for the `-acquire` and `-release` parts. Not sure if worth it, but it'd probably reduce clutter (one root span for acquire, release, use [which will supposedly contain spans of its own, created by the user])

PR #20, opened 2019-09-16
Scala 2.13 support

iRevive:
This PR adds the support of Scala 2.13.


.travis.yml
```diff
@@ -15,15 +15,16 @@ jobs:
   include:
     # stage="test" if no stage is specified
     - env: TEST="build"
-      script: sbt headerCheck compile test
+      script: sbt ";headerCheck; + compile; + test"
     # run ci-release only if previous stages passed
     - stage: release
-      script: sbt ci-release
+      script: sbt + ci-release
```
tpolecat: `ci-release` doesn't need `+`... it's a command that calls `+publishSigned` and so on.

build.sbt
```diff
@@ -1,3 +1,7 @@
+lazy val scala211Version = "2.11.12"
+lazy val scala212Version = "2.12.10"
+lazy val scala213Version = "2.13.0"
```
tpolecat: 2.13.1

.travis.yml
iRevive: Good point. Done.

build.sbt
iRevive: Currently, kind-projector doesn't support 2.13.1. 
Issue: https://github.com/typelevel/kind-projector/issues/116.

Should we wait for their release first?

PR #47, opened 2019-11-12
Semigroup/monoid instances

janstenpickle:
I've add some instances for kernel, span, and entrypoint. The motivation for doing so is to allow combination of different tracers, for example you may want to combine the logger with lightstep.

I haven't added laws yet, but am happy to do so if you like the idea of doing things this way.

### DIFF
--- a/modules/core/src/main/scala/Kernel.scala
+++ b/modules/core/src/main/scala/Kernel.scala
@@ -4,9 +4,20 @@
 
 package natchez
 
+import cats.kernel.Monoid
+
 /**
  * An opaque hunk of data that can we can hand off to another system (in the form of HTTP headers),
  * which can then create new spans as children of this one. By this mechanism we allow our trace
  * to span remote calls.
  */
 final case class Kernel(toHeaders: Map[String, String])
+
+object Kernel {
+  implicit def kernelMonoid: Monoid[Kernel] =
+    new Monoid[Kernel] {
+      override def empty: Kernel = Kernel(Map.empty)
+
+      override def combine(x: Kernel, y: Kernel): Kernel = Kernel(x.toHeaders ++ y.toHeaders)
+    }
+}
--- a/modules/core/src/main/scala/Kernel.scala
+++ b/modules/core/src/main/scala/Kernel.scala
@@ -5,19 +5,33 @@
 package natchez
 
 import cats.kernel.Monoid
+import natchez.Kernel.HeaderKey
 
 /**
- * An opaque hunk of data that can we can hand off to another system (in the form of HTTP headers),
- * which can then create new spans as children of this one. By this mechanism we allow our trace
- * to span remote calls.
- */
-final case class Kernel(toHeaders: Map[String, String])
+  * An opaque hunk of data that can we can hand off to another system (in the form of HTTP headers),
+  * which can then create new spans as children of this one. By this mechanism we allow our trace
+  * to span remote calls.
+  */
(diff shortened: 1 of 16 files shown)


tpolecat:
I like this idea but I'm a little weirded out by the `Kernel` monoid. Let me think about it a little bit.

janstenpickle:
Thanks Rob.

The kernel monoid got me too. It does potentially lose information when combining kernels genereated from different tracers. I did think about having a second type of kernel that could be namespaced by tracer, but that means the user would have to work out what to do with the different kernels.

Another approach could be that the tracer implementation decides what to do when combining with existing kernel. This could potentially break span's associativity though.

janstenpickle:
@tpolecat I've had a go a differentiating headers based on their backend. I opted to tag each header key by its backend, therefore when a certain backend is serialising a kernel to headers it can pick out only the values directly applicable to its implementation. What do you think?

janstenpickle:
@tpolecat sorry to badger you, but I'd love to get your thoughts on the changes to kernel

modules/core/src/main/scala/Kernel.scala
```diff
@@ -4,9 +4,34 @@
...
- * An opaque hunk of data that can we can hand off to another system (in the form of HTTP headers),
- * which can then create new spans as children of this one. By this mechanism we allow our trace
- * to span remote calls.
- */
-final case class Kernel(toHeaders: Map[String, String])
+  * An opaque hunk of data that can we can hand off to another system (in the form of HTTP headers),
+  * which can then create new spans as children of this one. By this mechanism we allow our trace
+  * to span remote calls.
+  */
+final case class Kernel(keyedHeaders: Map[HeaderKey, String]) {
```
tpolecat: We're assuming here that `HeaderKey` (an open trait) has sane universal equality and hashing, which I am reluctant to do. I wonder if it would work better to have `case class Provider(name: String)` and then `Map[Provider, Map[String, String]]`.

I think we may also want to prefix the provider to the header keys because we can still have the problem where two providers use the same key and it means different things (and one of them will be clobbered).

janstenpickle: Thanks for taking a look!

> We're assuming here that HeaderKey (an open trait) has sane universal equality and hashing, which I am reluctant to do

Ah yeah, I didn't think about this, bad times!

> I think we may also want to prefix the provider to the header keys because we can still have the problem where two providers use the same key and it means different things (and one of them will be clobbered).

I suppose we only want to do this when converting the keys from `Map[Provider, Map[String, String]]` to `Map[String, String]`, the various providers should be looking up their specific map in the map of maps.

tpolecat: Right, each provider sees its own map. This still raises the possibility of concatenating two spans from the same provider, in which case the one on the left will win if there are key collisions. This is ok from a semigroup point of view but it will lead to puzzling behavior (although it will be hard to provoke)

janstenpickle: Ah yeah, unless we are able to assign some sort of identifer to the provider on creation which will ensure the headers are always mapped back correctly. Of course this only works when encoding headers and not decoding. The only way around that may be to have separate types for when continuing in an entry point and propagating context from a span.

janstenpickle:
I'm finally coming back to this after months... sorry! 😅

As you've probably seen I've made my [own tracing implementation](https://github.com/janstenpickle/trace4cats) which allows exporters to be comined with a monoid without having to worry about overlapping headers, so am going to close this.

Thanks for the feedback!

PR #53, opened 2019-11-24
Remove Skunk instructions from README

valencik:
These instructions are no longer necessary after https://github.com/tpolecat/natchez/pull/48


tpolecat:
Thanks!

PR #69, opened 2020-01-03
OpenCensus consistent bracket on root spans as well as child spans

david-cromberge:
Previously, only child spans inspected the exit case for a span
resource to accurately report errors.  This change ensures that
spans created from a kernel or root follow the same approach.


tpolecat:
Thanks!

PR #72, opened 2020-01-15
Add Trace#spanR

kubukoz:
See #73 for motivation.

This is a very quick initial idea, and it needs a lot of closer looks (e.g. to loosen the constraints, as Sync might not be needed wherever it's used). But I experimented with it locally, and it seems to work (e.g. with jaeger).

### DIFF
--- a/modules/core/src/main/scala/Span.scala
+++ b/modules/core/src/main/scala/Span.scala
@@ -5,10 +5,6 @@
 package natchez
 
 import cats.effect.Resource
-import cats.effect.Bracket
-import cats.Defer
-import cats.Applicative
-import cats.~>
 
 /** An span that can be passed around and used to create child spans. */
 trait Span[F[_]] { self =>
@@ -24,10 +20,4 @@ trait Span[F[_]] { self =>
 
   /** Resource that yields a child span with the given name. */
   def span(name: String): Resource[F, Span[F]]
-
-  def mapK[G[_]](fk: F ~> G)(implicit b: Bracket[F, Throwable], deferG: Defer[G], applicativeg: Applicative[G]): Span[G] = new Span[G] {
-    def put(fields: (String, TraceValue)*): G[Unit] = fk(self.put(fields:_*))
-    def kernel: G[Kernel] = fk(self.kernel)
-    def span(name: String): Resource[G,Span[G]] = self.span(name).mapK(fk).map(_.mapK(fk))
-  }
 }
--- a/modules/core/src/main/scala/Span.scala
+++ b/modules/core/src/main/scala/Span.scala
@@ -5,9 +5,13 @@
 package natchez
 
 import cats.effect.Resource
+import cats.effect.Bracket
+import cats.Defer
+import cats.Applicative
+import cats.~>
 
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/Span.scala
```diff
@@ -21,4 +25,9 @@ trait Span[F[_]] {
   /** Resource that yields a child span with the given name. */
   def span(name: String): Resource[F, Span[F]]
 
+  def mapK[G[_]](fk: F ~> G)(implicit b: Bracket[F, Throwable], deferG: Defer[G], applicativeg: Applicative[G]): Span[G] = new Span[G] {
+    def put(fields: (String, TraceValue)*): G[Unit] = fk(self.put(fields:_*))
```
kubukoz: I think `put` could take a List of fields instead, with a varargs overload implemented as an alias...

modules/core/src/main/scala/Trace.scala
```diff
@@ -68,8 +78,10 @@ object Trace {
       Kleisli(_.put(fields: _*))
 
     def span[A](name: String)(k: Kleisli[F, Span[F], A]): Kleisli[F,Span[F],A] =
```
kubukoz: As for naming: maybe this should be `spanned` and `spanR` could be the new `span`?

modules/core/src/main/scala/Trace.scala
```diff
@@ -24,6 +24,7 @@ trait Trace[F[_]] {
   /** Create a new span, and within it run the continuation `k`. */
   def span[A](name: String)(k: F[A]): F[A]
 
+  def spanR(name: String): Resource[F, Span[F]]
```
tpolecat: The idea with `Trace` is that there is an ambient span, so it would be more consistent to have the resource yield `Unit` I think.

kubukoz: That makes it easier... And makes sense, as I can put my stuff into the span by using the method on Trace.

modules/core/src/main/scala/Trace.scala
kubukoz: btw. @tpolecat what do you think about having this as syntax on `F[A]`?

tpolecat: Where `fa.span("foo")` = `Trace[F].span("foo")(fa)`? Seems useful.

kubukoz: Yeah, just like that. I'll give a stab at these in different PRs then.

kubukoz:
Alright, so I tried to actually use this, and it doesn't work. The resource gives the span a lifetime, sure, but the span doesn't actually end up "the" span in the environment. It just doesn't seem possible to simulate `Kleisli.local(_ => newSpan)(...)` with Resource. I'll keep trying but for now the shape of Resource stands in the way.

kubukoz:
I didn't figure anything out that would make this work with Kleisli. I'm pretty sure the encoding of Resource will be the blocker here. @tpolecat if you have a better idea, I'm all ears, but for now I'll close this :)

PR #74, opened 2020-01-16
add log module to build aggregate

bpholt:
I noticed the log module wasn't being published, but I found it useful while I was playing with the library. I think with these changes, it should be published during the next release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -49,8 +49,8 @@ lazy val natchez = project
     crossScalaVersions := Nil,
     publish / skip     := true
   )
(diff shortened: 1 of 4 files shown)


tpolecat:
What's the intent of `MapValuesToJson`?

bpholt:
I was getting deprecation warnings about `mapValues` in Scala 2.13, so I wanted to present a common API that could cross-compiled without warnings.

kubukoz:
oh no, I didn't see this 😢 @bpholt if you don't mind, my solution to the `mapValues` problem involved less code (and zero version-specific code), so maybe let's merge that instead? #74

(also, circe 0.12 got old real quick)

bpholt:
I don’t mind at all! I’ll close this one.

PR #75, opened 2020-01-16
Http4s Middlewares for Client and Server

tbrown1979:
Addresses #5 

Okay, so there's been a lot of talk on #5 about what we should do for an http4s module. I've had these changes locally for a _long_ time so I've decided to PR it. They're similar to what I've done for an internal library at my company. I realize this PR isn't really inline with what has been suggested on #5, so feel free to request changes and I will happily make them. I just wanted to get something up that we can iterate on and ultimately get merged.


tpolecat:
I need to reconcile this with something I wrote myself, so hang tight. Thanks.

modules/http4s/client/src/main/scala/Middleware.scala
```diff
@@ -0,0 +1,28 @@
...
+import natchez.Tags._
+import org.http4s.client.Client
+import org.http4s.{Header, Headers}
+
+object Middleware {
+
+  def spannedClient[F[_]](client: Client[F], clientName: String)(implicit F: Sync[F], S: Span[F]): Client[F] = {
+    Client[F](request =>
+      for {
+        s      <- S.span(s"http4s-request-$clientName")
```
kubukoz: I was under the impression Span isn't meant to be used like that (it's a data type and not a type class) and that `Trace[F].span` / `Trace[F].add` should be used to create new spans. Is that right?

tbrown1979: Ahh, I suppose the difference between `Trace` and `Span` is just not very clear to me. I can change this to use `Trace` instead if that's desired!

tbrown1979:
I have some much better changes to make here I think, so I'll update this sometime in the next few days.

DamienOReilly:
Any new updates here?

tbrown1979:
@DamienOReilly Sorry, haven't been able to swing back around. I'll see if I can get this updated soon!

DamienOReilly:
@tbrown1979 no worries, thanks for the update.

tdbgamer:
@tbrown1979 Is there any way someone else could assist with the completion of this PR? I'm making some http4s services soon and really want to use this library for tracing. I'm willing to contribute and am really trying to avoid forking this and merging it myself as a temporary measure :sweat_smile:

tbrown1979:
I won't be finishing this PR. If someone else would like to pick this up feel free!

PR #79, opened 2020-01-18
Refactor out a generic OpenTracing span/tracer

bpholt:
This should let us more easily add OpenTracing backends other than Jaeger. 

For example, I have it working with https://github.com/opentracing-contrib/java-xray-tracer locally, but that project is apparently having some problems with publishing artifacts. Once there's a reliable artifact available I'll PR the XRay backend too.

### DIFF
--- a/modules/opentracing/src/main/scala/OTSpan.scala
+++ b/modules/opentracing/src/main/scala/OTSpan.scala
@@ -36,6 +36,9 @@ private[opentracing] final case class OTSpan[F[_] : Sync](tracer: ot.Tracer,
     }
 
   def span(name: String): Resource[F, Span[F]] =
-    makeSpan(tracer)(Sync[F].delay(tracer.buildSpan(name).asChildOf(span).start))
-
+    makeSpan(tracer)(Sync[F].delay {
+      val newSpan: ot.Span = tracer.buildSpan(name).asChildOf(span).start()
+      tracer.scopeManager().activate(newSpan)
+      newSpan
+    })
 }
--- a/modules/opentracing/src/main/scala/OTSpan.scala
+++ b/modules/opentracing/src/main/scala/OTSpan.scala
@@ -3,21 +3,18 @@
 // For more information see LICENSE or https://opensource.org/licenses/MIT
 
 package natchez
-package jaeger
+package opentracing
 
-import io.{ opentracing => ot }
-import cats.effect.Sync
-import cats.effect.Resource
+import cats.effect.{Resource, Sync}
 import cats.implicits._
-import io.opentracing.propagation.Format
-import io.opentracing.propagation.TextMapAdapter
+import io.opentracing.propagation.{Format, TextMapAdapter}
+import io.{opentracing => ot}
+import natchez.opentracing.OTTracer.makeSpan
 
 import scala.jdk.CollectionConverters._
 
-private[jaeger] final case class JaegerSpan[F[_]: Sync](
-  tracer: ot.Tracer,
-  span:   ot.Span
-) extends Span[F] {
+private[opentracing] final case class OTSpan[F[_] : Sync](tracer: ot.Tracer,
+                                                          span: ot.Span) extends Span[F] {
   import TraceValue._
 
   def kernel: F[Kernel] =
@@ -33,15 +30,12 @@ private[jaeger] final case class JaegerSpan[F[_]: Sync](
 
   def put(fields: (String, TraceValue)*): F[Unit] =
     fields.toList.traverse_ {
-      case (k, StringValue(v))  => Sync[F].delay(span.setTag(k, v))
-      case (k, NumberValue(v))  => Sync[F].delay(span.setTag(k, v))
+      case (k, StringValue(v)) => Sync[F].delay(span.setTag(k, v))
+      case (k, NumberValue(v)) => Sync[F].delay(span.setTag(k, v))
       case (k, BooleanValue(v)) => Sync[F].delay(span.setTag(k, v))
     }
 
-  def span(name: String): Resource[F,Span[F]] =
-    Resource.make(
-      Sync[F].delay(tracer.buildSpan(name).asChildOf(span).start))(
-      s => Sync[F].delay(s.finish)
-    ).map(JaegerSpan(tracer, _))
+  def span(name: String): Resource[F, Span[F]] =
+    makeSpan(tracer)(Sync[F].delay(tracer.buildSpan(name).asChildOf(span).start))
 
 }
--- a/modules/opentracing/src/main/scala/OTTracer.scala
+++ b/modules/opentracing/src/main/scala/OTTracer.scala
@@ -1,0 +1,70 @@
+// Copyright (c) 2019 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package natchez
+package opentracing
+
+import cats.Applicative
+import cats.effect.{ExitCase, Resource, Sync}
+import cats.syntax.applicative._
+import cats.syntax.functor._
+import cats.syntax.flatMap._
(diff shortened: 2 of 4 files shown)


modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,54 @@
...
+import io.{opentracing => ot}
+import io.opentracing.propagation.{Format, TextMapAdapter}
+
+import scala.jdk.CollectionConverters._
+
+object OTTracer {
+
+  private[opentracing] def finishSpan[F[_] : Sync](t: ot.Tracer): ot.Span => F[Unit] = s =>
+    Sync[F].delay {
+      t.scopeManager().activate(s)
```
bpholt: I had to add this to get it working with XRay (which uses a ThreadLocal to track the active span in its SDK—afaict that can mostly be ignored, but this sets the span in the ThreadLocal before it's finished on the next line)

Full disclosure: I haven't tested it with Jaeger, but from some of the GitHub threads I was reading, it sounds like OpenTracing implementations are meant to use ScopeManager for this purpose anyway, so I'm hopeful it will be fine.

modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,54 @@
...
+
+          Resource.make(acquireSpan)(finishSpan(t)).map(OTSpan(t, _))
+        }
+
+        override def continueOrElseRoot(name: String, kernel: Kernel): Resource[F, Span[F]] =
+          continue(name, kernel) flatMap {
+            case null => root(name) // hurr, means headers are incomplete or invalid
+            case a    => a.pure[Resource[F, *]]
+          } recoverWith {
+            case _: Exception => root(name)
```
bpholt: This was a more specific exception in the Jaeger implementation. I'm not sure if widening it is a problem—if it is, I can let implementations pass in an exception handler (or something like that).

bpholt:
I'm not sure what's up with the build error—it [worked fine on my branch](https://travis-ci.org/bpholt/natchez/builds/638878149). Maybe you can restart the build and let it try again?

tpolecat:
The problem before was that different back ends used different, incompatible versions of OpenTracing. I will have a look when I can. I restarted the build.

bpholt:
The AWS Xray library used OpenTracing 0.31. I think Jaeger used 0.33 (which is currently the latest version, and maybe likely to remain that way since OpenTracing and OpenCensus merged to form OpenTelemetry). I will see if I can upgrade the Xray library to use 0.33.

bpholt:
I updated OpenTracing to 0.33, and added some span completion logic to make sure exceptions are attached if a span continuation fails.

modules/opentracing/src/main/scala/OTSpan.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def put(fields: (String, TraceValue)*): F[Unit] =
+    fields.toList.traverse_ {
+      case (k, StringValue(v)) => Sync[F].delay(span.setTag(k, v))
+      case (k, NumberValue(v)) => Sync[F].delay(span.setTag(k, v))
+      case (k, BooleanValue(v)) => Sync[F].delay(span.setTag(k, v))
+    }
+
+  def span(name: String): Resource[F, Span[F]] =
+    makeSpan(tracer)(Sync[F].delay {
```
kubukoz: WDYT about adding an optional `parent` parameter to `makeSpan` and putting the common effectful part inside `makeSpan`?

modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,77 @@
...
+import io.{opentracing => ot}
+
+import scala.jdk.CollectionConverters._
+
+object OTTracer {
+
+  private [opentracing] def makeSpan[F[_] : Sync](t: ot.Tracer)(acquireSpan: F[ot.Span]): Resource[F, Span[F]] =
+    Resource.makeCase(acquireSpan)(finishSpan(t)).map(OTSpan(t, _))
+
+  private def finishSpan[F[_] : Bracket[*[_], Throwable] : Sync](t: ot.Tracer): (ot.Span, ExitCase[Throwable]) => F[Unit] =
```
kubukoz: Sync includes this bracket already

modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,77 @@
...
+              newSpan
+            }
+          })
+        }
+
+        override def continueOrElseRoot(name: String, kernel: Kernel): Resource[F, Span[F]] =
+          continue(name, kernel) flatMap {
+            case null => root(name) // hurr, means headers are incomplete or invalid
+            case a    => a.pure[Resource[F, *]]
+          } recoverWith {
```
kubukoz: this would be `orElse(root(name))` I think

modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,77 @@
...
+      def attachPossibleException: ExitCase[Throwable] => F[Unit] = {
+        case ExitCase.Error(ex) =>
+          val map: java.util.Map[String, Throwable] = new java.util.HashMap[String, Throwable]
+          map.put(Fields.ERROR_OBJECT, ex)
+
+          Sync[F].delay(s.log(map)).void
+        case _ => Applicative[F].unit
+      }
+
+      Bracket[F, Throwable].guarantee(attachPossibleException(exitCase))(Sync[F].delay {
```
kubukoz: There's syntax for `guarantee` in `cats.effect.syntax.bracket._`. Also, I think `a.attempt *> b` would be enough as `finishSpan` will run in resource cleanup, which is uncancelable.

modules/opentracing/src/main/scala/OTTracer.scala
```diff
@@ -0,0 +1,77 @@
...
+              val newSpan: ot.Span = t.buildSpan(name).asChildOf(context).start()
+              t.scopeManager().activate(newSpan)
+              newSpan
+            }
+          })
+        }
+
+        override def continueOrElseRoot(name: String, kernel: Kernel): Resource[F, Span[F]] =
+          continue(name, kernel) flatMap {
+            case null => root(name) // hurr, means headers are incomplete or invalid
```
kubukoz: I have a feeling this null doesn't do what it was originally meant to do. `newSpan`, the `ot.Span` instance might be null, but the `Span[F]` instance most definitely isn't (the null will just be in that `OTSpan`'s `span` field).

I also feel weird about `continue` potentially returning an invalid span, but that's a matter for a different discussion.

modules/opentracing/src/main/scala/OTSpan.scala
```diff
@@ -0,0 +1,44 @@
...
+import io.{opentracing => ot}
+import natchez.opentracing.OTTracer.makeSpan
+
+import scala.jdk.CollectionConverters._
+
+private[opentracing] final case class OTSpan[F[_] : Sync](tracer: ot.Tracer,
+                                                          span: ot.Span) extends Span[F] {
+  import TraceValue._
+
+  def kernel: F[Kernel] =
```
kubukoz: Defs without parameters can be vals. As a bonus, we get fewer thunk allocations in `delay` (reusing the same thunk for the lifetime of this OTSpan instance)

kubukoz:
Hey @bpholt, do you want to continue with this or can I pick it up from what you've made?

bpholt:
@kubukoz please go ahead! I think the only real changes I made when refactoring from the original Jaeger implementation were called out in my comments above, so most of this is really Rob's work anyway 🙂

tpolecat:
I can take this PR if you'd like to fix the conflicts but it wasn't immediately trivial. Sorry for sitting on this for so long, trying to get caught up. Please open a new PR if you'd like to continue.

PR #97, opened 2020-03-01
Add log module to root project aggregation

kubukoz:


kubukoz:
oof, log4cats isn't published for Scala 2.11 anymore. Do you still want to support that version, @tpolecat?

tpolecat:
Eh, just remove 2.11

kubukoz:
Yes sir :)

kubukoz:
Now the log module complains about Circe. I'll update that as well (0.11 isn't released for 2.13)

kubukoz:
Good to go on my side.

PR #100, opened 2020-03-12
Add scalafmt config (based on cats)

kubukoz:


kubukoz:
@tpolecat do you have any suggestions for this?

kubukoz:
@tpolecat do you have a config I could use instead of this? I was basing on cats, but if I remember correctly you didn't like what Cats does ;)

PR #112, opened 2020-04-03
sfl4j-jdk14 not necessary for natchez-jaeger

tbrown1979:


tpolecat:
Looks like already did this one. Thanks!

PR #119, opened 2020-04-20
Ensure non-null OpenCensus string attribute values

davecromberge:
Handling the error case for an effect attempts to record the underlying
exception's message as well as stack trace as span attributes.
Some exception types such as `java.net.ConnectException` do not have an
exception message, which used to result in null values for the `error.message`
attribute value.

OpenCensus does not permit null values to be recorded as the value
for a string attribute.  This would result in the error stack trace
being polluted with an additional top-level OpenCensus constructor
exception, obscuring the underlying issue.

### DIFF
--- a/modules/opencensus/src/main/scala/OpenCensusSpan.scala
+++ b/modules/opencensus/src/main/scala/OpenCensusSpan.scala
@@ -27,8 +27,10 @@ private[opencensus] final case class OpenCensusSpan[F[_]: Sync](
   override def put(fields: (String, TraceValue)*): F[Unit] =
     fields.toList.traverse_ {
       case (k, StringValue(v)) =>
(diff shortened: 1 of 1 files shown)


tpolecat:
Good catch, thanks! I'll take a PR to fix that if anyone wants to do it.

davecromberge:
@tpolecat Thanks for the quick response.  I introduced this behaviour in an earlier PR, and this PR attempts to fix the problem.  Perhaps you are confusing this with an issue?

tpolecat:
Haha, sorry. I obviously wasn't paying attention. Thanks!

tpolecat:
Travis isn't reporting status anymore but it passed. Need to switch to GH Actions.

PR #145, opened 2020-08-06
Add No-Op open tracing implementation

sbly:
This is a No-Op Open Tracing implementation that I've been using for testing


build.sbt
```diff
@@ -157,6 +157,18 @@ lazy val log = project
     )
   )
 
+lazy val noop = project
```
tpolecat, line 160: Can you add this to the `dependsOn` and `aggregate` for the root project?

sbly:
@tpolecat Thanks for reviewing - I made the requested change

tpolecat:
Please run `sbt headerCreate` to add file headers, then I think you're all set.

sbly:
Done!

tpolecat:
Thanks!

PR #153, opened 2020-09-02
Add Mock open tracing implementation

sbly:
Add a module that uses `opentracing-mock` that can be used for testing. 

I chose to just expose the `MockTracer` as a public member of the entrypoint, so we can access all its methods. LMK if you would prefer to access it only through public methods of the `MockEntrypoint` class


tpolecat:
Sure, why not? 😄 Thanks!

PR #165, opened 2020-09-25
Trace instances for monad transformers

bplommer:


tpolecat:
This looks great, thanks! Could you add a compilation test to ensure that the instances for `Trace[Kleisli[F, Span[F]` and the one for `Trace[Kleisl[F: Trace` aren't in conflict? i.e., that they are both accessible?

kubukoz:
Just a suggestion, cats-mtl now contains a `MonadPartialOrder` typeclass that should be useful for lifting. It's an extra dependency, but maybe it's worth considering? Maybe as a separate module away from core?

bplommer:
> Just a suggestion, cats-mtl now contains a `MonadPartialOrder` typeclass that should be useful for lifting. It's an extra dependency, but maybe it's worth considering? Maybe as a separate module away from core?

Interesting - I have an open PR for an mtl module (https://github.com/tpolecat/natchez/pull/166), so could be worth adding there.

tpolecat:
subsumed by #248

PR #170, opened 2020-10-05
Generalize the handling of Fields

rzeigler:
Span now has putErrorFields of Resource[F, Span[F]] => Resource[F, Span[F]].
The resulting resource proxies the input span and adds makeCase logic to attach data from errors with Fields.


tpolecat:
subsumed by #250

PR #193, opened 2020-11-11
Add scalajs support

pauljamescleary:
For many applications, the root of a trace starts in the browser on the client.
It is valuable to track the root back to the client application, which may make
several server side requests as part of the same user gesture.

In addition, the transit time for large payloads, as well as the scalajs
processing time to unpack and work with larger payloads, can have a significant
impact on overall performance from a user experience perspective.  As such, it
is extremely valuable to understand performance from a user/browser perspective
in order to find slow spots and hot spots in an application.

I have another PR that for log4cats, that adds scalajs logging allowing logs to
be sent to an HTTP4S endpoint.  Using natchez with log4cats client side will
allow tidy client side trace support, logging along side other natchez /
log4cats logs server side.

This PR does little more than move files around, modify the build to support
crossProject / scalajs for `core` and `log`.


tpolecat:
Wow I'm kind of amazed this compiles along with the dotty stuff. I'll merge and tag a release and see it blows up.

Thanks!

tpolecat:
I tagged `v0.0.14-M3` so give it a few minutes and see what happens.

pauljamescleary:
@tpolecat does not release JS support for dotty as there are a number of missing dependencies.  Only does JS release for cross versions 2.x which should be fine.

Thanks for the merge!

I will be pulling this all into an inhouse project, may wind up with some other PRs depending on how it shakes out, but the bulk of what I have now is in log4cats

PR #196, opened 2020-11-13
Fix crossProject support

pauljamescleary:
Classes were not being packaged properly for natchez-log module.  This was
discovered after doing local testing. 

Need to make sure that the `-2` directory is added to sources so they are released.

natchez-log for 0.0.14-M3 did not contain any classes.


tpolecat:
Can you explain this a little bit more? All the projects have `scala-2` and `scala-3` directories, added higher up in the build. Why do the js projects need special handling?

pauljamescleary:
@tpolecat the scalajs projects have `shared/src/main`, the `shared` directory is "special" to `crossProject` projects.  By default, it picks up `shared/src/main`, but one needs to override it to pick up source directories for scala2 vs. scala3.

I did some poking around in other projects, and sbt-spiewak seems to take care of this very well, so I dug into the Plugin and found this - https://github.com/djspiewak/sbt-spiewak/blob/7048b251e96773b6ebdc1100b243608628326351/core/src/main/scala/sbtspiewak/SpiewakPlugin.scala#L458

Following this into `crossProject` - the `sharedSrcDir` is set based on the cross project `CrossType`, which by default is `Full` in order to support cross compiling to scalajs and scala (as well as native if so desired) - https://github.com/portable-scala/sbt-crossproject/blob/e1ec68f1321a313abeb27c7bb7f7bca1a5ac9d91/sbt-crossproject/src/main/scala/sbtcrossproject/CrossType.scala#L62

**tl;dr** have to explicitly add cross compile options for custom source directories `scala-2` and `scala-3` when using `crossProject`

pauljamescleary:
@tpolecat wanted to bump this to fix the "natchez-log artifact empty" issue opened #208

tpolecat:
Sorry for the delay on this. I'm not sure what the issue with resolution is. Can you ensure that the sbt-dotty plugin is updated to latest?

pauljamescleary:
@tpolecat I will make sure to update the dotty plugin

tpolecat:
subsumed by #246

PR #213, opened 2020-12-02
DataDog - Log Errors

etspaceman:
This PR adds some error-handling to the Span creation, ensuring that errors are propagated to the tracer.


etspaceman:
@tpolecat any thoughts on this one? We use this currently at work and its a small blocker for us adopting the Datadog module directly.

PR #240, opened 2021-01-18
adding spanId to spans, traceId implementation for all OT backends

patseev:
This PR adds a `spanId` field to spans. This is useful for backends that provide search through spans and allow to correlate spans and logs (e.g. Datadog).

Also, for all OT-based backends `spanId` and `traceId` are being extracted from ot's `SpanContext` now, this allows us not to extract it from backend-specific headers (like it was implemented for Jaeger), SpanContext does this job for us.

I've tested this only for Jaeger and Datadog, but other implementations didn't have any implementation for traceId and spanId anyway.

OT doc states that `spanId` and `traceId` can be empty, in this case I consider it to be non-existent and return None


tpolecat:
Thanks!

modules/honeycomb/src/main/scala/HoneycombSpan.scala
```diff
@@ -41,8 +41,14 @@ private[honeycomb] final case class HoneycombSpan[F[_]: Sync](
...
     Resource.makeCase(HoneycombSpan.child(this, label))(HoneycombSpan.finish[F]).widen
 
-  def traceId: F[Option[String]] = traceUUID.toString.some.pure[F]
-  def traceUri: F[Option[URI]]   = none.pure[F] // TODO
+  def traceId: F[Option[String]] =
+    traceUUID.toString.some.pure[F]
+
+  def spanId: F[Option[String]] =
+    spanUUID.toString.some.pure[F]
+
```
tpolecat, line 49: Wait, how did this compile?

tpolecat: Ah the build is broken.

PR #241, opened 2021-01-20
Move code to reflect the package structure

onedigit:
Minor changes that move code in `core` module to a directory to reflect the package structure. Similar change in `examples`.


onedigit:
There are more places where the same change is needed

PR #255, opened 2021-01-29
add traceUri support for Datadog

patseev:
Added traceUri support to Datadog module.
Now creation of entryPoint requires a baseUri (because uris are different for EU and US DD accounts) but it's optional for backward compatability.


tpolecat:
Thanks for this, sorry for the delay!

PR #272, opened 2021-02-21
Update: cats effect, scalajs, collection compat, scala 3.0.0-RC1

keynmol:
* Drop Scala 3.0.0-M2
* Add Scala 3.0.0-RC1

Which lead to

* Update cats
* Update cats-effect to RC2
* Update collection-compat ([dropped](https://github.com/scala/scala-collection-compat/releases/tag/v2.4.2) support for M2)
* Update cats-mtl
* Update dotty plugin


tpolecat:
Thanks!

PR #283, opened 2021-03-02
Create entry points using the globally registered tracer

Jacoby6000:
This PR does a couple of things.

1. Adds class definitions for `EntryPoint`s that are based on opentracing apis, so they're no longer anonymous.
2. Makes those `EntryPoint`s public, allowing for initialization in unsafe contexts; necessary in something like a play project.  This should be safe, since the EntryPoints themselves are pure on initialization.
3. Allows users to initialize an EntryPoint using the  GlobalTracer from opentracing.
4. Makes natchez register the tracers created via the `entryPoint` methods to the GlobalTracer.

N.B.: This lets people shoot themselves in the foot if they want. If you have a java agent attached that creates a Datadog tracer, and then you use the global tracer to initialize a Jaeger tracer, that is valid code.  I can't say what will happen at runtime though.  I think a way around this is to have each module provide a method for creating an entry point via the global tracer.

I'm trying to maintain compat with this PR, so there's some weirdness we might need to work around. I'm happy to make any necessary changes to this PR, just let me know.  My team has found that initializing a new tracer while a java agent is attached can cause issues with modules like DataDog, because the DataDog agent attaches its own tracer, then blows up when you try to spawn a new one.


modules/opentracing/src/main/scala/natchez/opentracing/GlobalTracer.scala
```diff
@@ -0,0 +1,35 @@
...
+  def registerTracer[F[_]: Sync](tracer: ot.Tracer): F[Boolean] = Sync[F].delay(GT.registerIfAbsent(tracer))
+
+  /* Creates an entrypoint using the GlobalTracer, if a tracer has been registered.  This is particularly useful if you 
+   * are using an agent to initialize your tracer. 
+   *
+   * N.B. There is currently no way to determine what underlying tracer that the global tracer is using.
+   */
+   
+  def createEntryPoint[F[_]: Sync](makeEntryPoint: ot.Tracer => F[EntryPoint[F]]): F[Option[EntryPoint[F]]] =
+    fetch[F].flatMap(_.traverse(makeEntryPoint))
```
Jacoby6000, line 32: This is the new method that allows users to spawn a new EntryPoint instance using the GlobalTracer

Jacoby6000:
I went ahead and cleaned up the apis a bit.  Much more tolerable to look at and use now.

kubukoz:
> Makes those EntryPoints public, allowing for unsafe initialization; necessary in something like a play project.

I generally prefer doing the opposite, using `Resource#allocated` to unsafely register the resource in Play's lifecycle (including cleanup!). We did this a lot in $previousJob and it worked quite well. @majk-p any chance you could paste the resource allocator thing we used in Play?

modules/datadog/src/main/scala/DDEntryPoint.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import cats.effect._
+import cats.syntax.all._
+import io.opentracing.propagation.{Format, TextMapAdapter}
+import io.{opentracing => ot}
+
+import java.net.URI
+import scala.jdk.CollectionConverters._
+
+class DDEntryPoint[F[_]: Sync](tracer: ot.Tracer, uriPrefix: Option[URI]) extends EntryPoint[F] {
```
kubukoz, line 16: Can we make this final?

Jacoby6000:
TIL about `Resorce#allocated` :)

I should've written that as "allows initialization of the `EntryPoint`s in unsafe contexts, since the creating the EntryPoints themselves is pure.

I'll go back through and make the classes final.  As far as I know there is no other setup to worry about with the global tracer.  The rationale as far as I understand is that whatever registered itself to the global tracer will be managing the resources for the tracer's lifecycle.

majk-p:
@kubukoz did you mean this one:
```scala
trait ResourceAllocator[A] {
  def allocateWithStopHook(resource: Resource[IO, A], lifecycle: ApplicationLifecycle): A
}

object ResourceAllocator {

  def instance[A]: ResourceAllocator[A] = (resource: Resource[IO, A], lifecycle: ApplicationLifecycle) => {
    val (allocatedResource, shutdown) = resource.allocated.unsafeRunSync()
    lifecycle.addStopHook(() => shutdown.unsafeToFuture())
    allocatedResource
  }
}
```

kubukoz:
More or less 😄

kubukoz (approved):

tpolecat:
A problem I ran into before was that Lightstep and Jaeger used different, incompatible versions of the OT libraries. But maybe that's not an issue anymore.

If you can update the collection-compat version (can you factor it out while you're at it, since it's repeated like 7 times?) this should be ready to merge. Thanks!

tpolecat:
Thanks!

PR #316, opened 2021-04-12
iotrace

tpolecat:
This implements `Trace[IO]` in terms of `IOLocal`.


tpolecat:
hm something's wrong. will follow up.

PR #344, opened 2021-05-15
Allow spans to be propagated outside monadic scope

bplommer:
This introduces `Portal[F]`, a type alias for `F ~> F`, and adds `def portal: F[Portal[F]]` to `Trace`. The motivation is allowing tracing to take place across asynchronous boundaries, as shown in the modified sorting example.

`Portal` is probably a bad name, I'm open to better ideas


bplommer:
Superseded by #345

PR #345, opened 2021-05-16
Expose current span + ability to run in span, add fs2 module

bplommer:
This requires `Trace` to expose a type member representing the type of `Span` it uses, which is a bit nasty, but it allows some nice extra functionality.

The sorting example is modified to show propagation of spans outside of monadic scope.

I'm closing #344, as this achieves the same thing in a more general way.

### DIFF
--- a/modules/core/shared/src/main/scala/NoopSpan.scala
+++ b/modules/core/shared/src/main/scala/NoopSpan.scala
@@ -1,0 +1,31 @@
+// Copyright (c) 2019-2020 by Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package natchez
+
+import cats._
+import cats.effect.Resource
+import cats.syntax.all._
+
+import java.net.URI
+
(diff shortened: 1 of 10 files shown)


modules/core/shared/src/main/scala/NoopSpan.scala
```diff
@@ -0,0 +1,31 @@
...
+    Resource.eval(NoopSpan[F]().pure[F])
+
+  // TODO
+  def traceId: F[Option[String]] = none.pure[F]
+
+  def spanId: F[Option[String]] = none.pure[F]
+
+  def traceUri: F[Option[URI]] = none.pure[F]
+
+}
```
bplommer, line 31: This is copied from the `noop` module. Maybe that whole module should be moved into the core module, since the core module now also has a no-op span?

pnadon:
@bplommer what is the status of this project? We would love to use this within our project, since we use fs2 extensively throughout our code.

bplommer:
> @bplommer what is the status of this project? We would love to use this within our project, since we use fs2 extensively throughout our code.

I'm not sure what you mean? This is just a pull request, I don't maintain the project.

PR #368, opened 2021-06-11
Add an `opentelemetry` module

andimiller:
This adds a `natchez-opentelemetry` module which allows span reporting via the `opentelemetry-java` project.

The `Utils` object contains a helper to turn the `OpenTelemetry` `CompletableResultCode` class into an `F[Unit]` given `Async[F]`, this is useful for implementing `Resource`s for clean shutdown logic.

`Shutdownable` is a little abstraction to unify all the interfaces that have a `shutdown(): CompletableResultCode` method in `OpenTelemetry`, since they have no common interface.

`OpenTelemetrySpan` and `OpenTelemetryEntryPoint` are heavily based on the `natchez-jaeger` versions, with tweaks to make them work with `OpenTelemetry`.

The `OpenTelemetry` object which end users should interact with has these methods:
* `lift` can be used to lift any `F[T]` where `T` is an `OpenTelemetry` class with a `shutdown` method into a `Resource[F, T]`, it asks for a name to provide a nice error message
* `entryPoint` is the main way to make an `EntryPoint` and has a boolean flag to allow the user to globally register the `OpenTelemetry` if that's helpful, this defaults to false.
* `globalEntryPoint` will use the globally registered `OpenTelemetry` to create an `EntryPoint`

Note that this is currently using `OpenTelemetry` libraries at `1.4.0-SNAPSHOT` because I found a bug while developing this this broke the shutdown logic.
The issue is here https://github.com/open-telemetry/opentelemetry-java/issues/3306 and it was closed by this PR https://github.com/open-telemetry/opentelemetry-java/pull/3307 so it should make it into the next release.


andimiller:
I can also port this over to the 0.1 branch pretty easily, we just needed it for 0.0

zmccoy:
I was just looking for this today!

modules/opentelemetry/src/main/scala/natchez/opentelemetry/OpenTelemetry.scala
```diff
@@ -0,0 +1,60 @@
...
+  // Helper methods to help you construct Otel resources that clean themselves up
+  // We need a name so the failure error can contain something useful
+  def lift[F[_]: Async, T: Shutdownable](name: String, create: F[T]): Resource[F, T] =
+    Resource.make(create) { t =>
+      Sync[F].delay { Shutdownable[T].shutdown(t) }
+        .flatMap(Utils.asyncFromCompletableResultCode(s"$name cleanup", _))
+    }
+
+  def entryPoint[F[_] : Sync](uriPrefix: Option[URI] = None, globallyRegister: Boolean = false)(
+    configure: OpenTelemetrySdkBuilder => Resource[F, OpenTelemetrySdkBuilder]
```
zmccoy, line 28: Would the signature for configure be more accurate to be `OpenTelemetrySdkBuilder => F[OpenTelemetrySdkBuilder]`?  Is there a configuration that you'd want to control a finalizer on it?

mpilquist:
I merged #539. Please take a look and see if there's anything from this PR we should incorporate.

PR #381, opened 2021-07-08
Add logging capabilities to Trace and Span

sbly:
Add the ability to log time-specific messages containing key-value pairs, as outlined by the OpenTracing spec: https://opentracing.io/docs/overview/tags-logs-baggage/#logs

I have tried to give each implementation of `Span` a sensible implementation. Some do not support logging in any reasonable manner, and others have a slightly different notion of logging


mpilquist:
Apologies on the delay on this one. Any interest in reviving it?

PR #427, opened 2021-09-27
X-Ray

christiankjaer:
Will serialize segments as described [here](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) and send them to the running X-Ray daemon as described [here](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-daemon)

Will not send child spans before the parent finishes.


bpholt:
@tpolecat When you get a chance, would you mind approving the GitHub Actions workflow for this PR? It'd be nice to get that feedback.

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -0,0 +1,272 @@
...
+
+  val Header = "X-Amzn-Trace-Id"
+
+  final case class XRayHeader(
+      traceId: String,
+      parentId: Option[String],
+      sampled: Boolean
+  ) {
+    def toKernel: Kernel =
+      Kernel(Map("X-Amzn-Trace-Id" -> s"Root=$traceId;Parent=$parentId;Sampled=${if (sampled) "1" else "0"}"))
```
armanbilge, line 153: Some code dupe with lines 120 and 145.

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -0,0 +1,272 @@
...
+    kernel.toHeaders
+      .get(Header)
+      .flatMap(parseHeader)
+      .map(x => fromHeader(name, x, entry))
+      .getOrElse(root(name, entry))
+
+  def root[F[_] : Concurrent : Clock : Random : Parallel](name: String,
+                                                          entry: XRayEntryPoint[F]): F[XRaySpan[F]] =
+    (segmentId[F], traceId[F], Clock[F].realTime, Ref[F].of(Map.empty[String, Json]), Ref[F].of(List.empty[JsonObject]))
+      .parMapN { (sId, tId, t, fields, children) =>
```
armanbilge, line 227: Out of curiosity any specific reason for the use of `Parallel` here and everywhere?

build.sbt
```diff
@@ -314,6 +314,26 @@ lazy val noop = crossProject(JSPlatform, JVMPlatform)
...
+  .in(file("modules/xray"))
+  .dependsOn(core)
+  .enablePlugins(AutomateHeaderPlugin)
+  .settings(commonSettings)
+  .settings(
+    name        := "natchez-xray",
+    description := "AWS X-Ray bindings implementation",
+    libraryDependencies ++= Seq(
+      "io.circe"          %%% "circe-core"      % "0.14.1",
+      "co.fs2"            %%% "fs2-io"          % "3.2.0",
```
armanbilge, line 327:
```suggestion
      "co.fs2"            %%% "fs2-io"          % "3.2.2",
```

armanbilge (commented):
Just some nits :)

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
christiankjaer: Maybe @bpholt can answer this.

tpolecat:
Since it looks like you're still working on this I'm going to change it to a draft. Click the thing when it's ready and it looks like we can get it merged quickly. Thanks!

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -0,0 +1,272 @@
...
+
+  def spanId: F[Option[String]] = segmentId.some.pure[F]
+
+  def traceUri: F[Option[URI]] = none[URI].pure[F]
+
+  /* The X-Ray documentation says to use microsecond resolution when available:
+   * https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-fields
+   */
+  private def toEpochSeconds(t: FiniteDuration): Double =
+    t.toMicros.toDouble / 1000000
```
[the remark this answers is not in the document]
christiankjaer: Hmm. Then I would actually like to bring back the Java stuff, since microsecond resolution would be rather nice for short segments.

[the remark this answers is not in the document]
armanbilge: That's fair, but if possible I'd prefer to fix this upstream in Cats Effect. It's a common problem, and a Java solution also won't work well with Scala.js.

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
bpholt: `Parallel` isn't strictly necessary; I just wanted to use `parMapN` instead of `mapN`.

armanbilge: Right, why is that? :) At a first glance I would think the overhead of `parMapN` exceeds any performance gain.

bpholt: That's definitely possible. I just typically default to `Parallel`. If it's not worth the overhead (or extra complexity) then I'm fine with removing it.

armanbilge: Yeah according to https://github.com/typelevel/cats-effect/pull/2239#issue-971957270 there's about any 85% overhead for using `Parallel`. If the ops we were parallelizing were async or CPU-bound it could make sense but I'm pretty sure these are al trivial. I would drop it :)

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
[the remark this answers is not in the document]
christiankjaer: If it's solvable, then I guess we keep the code as is, and then it will be a little sub-optimal until it's fixed in CE.

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -0,0 +1,272 @@
...
+              "stack" -> ex
+                .getStackTrace
+                .map(x =>
+                  Json.obj(
+                    "line" -> x.getLineNumber.asJson,
+                    "path" -> x.getFileName.asJson,
+                    "label" -> x.getMethodName.asJson
+                  )
+                )
+                .asJson
```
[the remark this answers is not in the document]
christiankjaer: Well. The current design has the child segments as `JsonObject`s. Doing this in a more typed way requires a bigger refactor.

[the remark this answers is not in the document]
armanbilge: I mean, it doesn't need to be anything fancy. It's like declaring it its own method? Just wrap it up in an `Encoder` :)

christiankjaer:
@armanbilge @bpholt Could one of you guys take a quick look again before I set this as ready?

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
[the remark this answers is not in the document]
christiankjaer: Isn't encoders implicitly encoding Json? Then I would need to convert it to a JsonObject again which might return `None`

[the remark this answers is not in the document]
armanbilge: https://circe.github.io/circe/api/io/circe/Encoder$$AsObject.html

armanbilge (changes requested):
Personally I'd like to see https://github.com/tpolecat/natchez/pull/427#discussion_r746561509 and https://github.com/tpolecat/natchez/pull/427#discussion_r746554360 addressed. But I'm not a maintainer so no stake here :)

modules/xray/shared/src/test/scala/XRaySuite.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  val rootId = "1-5759e988-bd862e3fe1be46a994272793"
+  val parentId = "53995c3f42cd8ad8"
+
+  val noParent = "Root=1-5759e988-bd862e3fe1be46a994272793;Sampled=1"
+  val parent =
+    "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1"
+  val notSampled = "Root=1-5759e988-bd862e3fe1be46a994272793;Sampled=0"
+
+  test("header parsing") {
```
armanbilge, line 20: Maybe overkill, but since you have both header parsing/encoding we could have a scalacheck prop checking the roundtrip.

christiankjaer: I thought about it. Came to the conclusion that it would be overkill.

bpholt: FWIW I also think it's a good idea, but IMO it can come in a followup PR if someone wants to do it. Similarly there are places where it might be more clear to use JSON literals instead of constructing JSON objects from key/value pairs, but I think it'd be fine to do that later too.

armanbilge (approved):
Looking good to me, thanks for addressing all my feedback! Apologies since we dragged this a bit out from your original scope, but much appreciated 😁

modules/xray/shared/src/test/scala/XRaySuite.scala
armanbilge: Yes I had the same thought about JSON literals :)

christiankjaer:
@armanbilge Right. To give some context, the original implementation was just a simple adaption of the logging implementation.

christiankjaer:
@tpolecat
I think the arguments are all over now. You can merge it if you think it's alright.
Thanks!

modules/xray/shared/src/main/scala/natchez/xray/XRayEnvironment.scala
```diff
@@ -0,0 +1,36 @@
...
+        OptionT(Sync[F].delay(env.get("AWS_XRAY_DAEMON_ADDRESS")))
+          .subflatMap(SocketAddress.fromStringIp)
+          .value
+
+      override def traceId: F[Option[String]] =
+        Sync[F].delay(env.get("_X_AMZN_TRACE_ID"))
+
+      override def kernelFromEnvironment: F[Kernel] =
+        OptionT(traceId)
+          .map(XRaySpan.XRayHeader(_, None, sampled = true).toKernel)
```
bpholt, line 33: I think I got this wrong in my pull request. It looks like `traceId` returns something like `Root=1-619446d4-2de198a61485ba3d50bf889d;Parent=7c143d692eca32b3;Sampled=1` if `_X_AMZN_TRACE_ID` is present in the environment, meaning the value should be parsed instead of just dropped into a new `XRayHeader` directly.

Instead, I think this works:

```scala
OptionT(traceId)
  .subflatMap(XRaySpan.parseHeader)
  .map(_.toKernel)
  .getOrElse(Kernel(Map.empty))
```

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -0,0 +1,309 @@
...
+  def fromKernel[F[_]: Concurrent: Clock: Random](
+      name: String,
+      kernel: Kernel,
+      entry: XRayEntryPoint[F]
+  ): F[XRaySpan[F]] =
+    kernel.toHeaders
+      .get(Header)
+      .flatMap(parseHeader)
+      .map(x => fromHeader(name, x, entry))
+      .get
```
bpholt, line 233: I think this `.get` is unsafe because it may throw an exception outside the scope of the `F[_]` effect. One option would be to do something like this:

```scala
cats.data.OptionT.fromOption[F](kernel.toHeaders.get(Header))
  .subflatMap(parseHeader)
  .semiflatMap(fromHeader(name, _, entry))
  .getOrElseF(new NoSuchElementException().raiseError)
```

modules/xray/shared/src/main/scala/natchez/xray/XRayEntryPoint.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  def make(span: F[XRaySpan[F]]): Resource[F, Span[F]] =
+    Resource.makeCase(span)(XRaySpan.finish(_, this, _)).widen
+
+  def root(name: String): Resource[F, Span[F]] =
+    make(XRaySpan.root(name, this))
+
+  def continue(name: String, kernel: Kernel): Resource[F, Span[F]] =
+    make(XRaySpan.fromKernel(name, kernel, this))
```
armanbilge, line 36: I've been thinking about this, just a random (bad) idea/question:
Besides the kernel provided by the user, should this also attempt to source a kernel from the environment as a fallback? Isn't that also some form of "continuing"?

bpholt: That's an interesting idea. Basically, look for the headers needed in the kernel provided by the user, and if they're missing, then check the environment? I think that could work.

armanbilge: Yes, I think this is worth pursuing, so that clients don't have to specialize to this backend by manually sourcing the kernel from the environment.

PR #435, opened 2021-10-14
Generic OpenTracing tracer

kubukoz:
This is mostly moving the code from the DataDog implementation, with a few small differences:

- the code that defines the trace URI is refactored out so it can be specified by the implementations, e.g. I have a different path for Zipkin
- in case of error, we don't do `.log` anymore but actually put some fields specifying that the span failed.

This is what a failed span looks like in DataDog with this code:

image

(parent spans also inherit the error if it's not caught with a `fa.attempt` or otherwise)

Previous attempt: https://github.com/tpolecat/natchez/pull/79

Some things I'm not so sure about:

- ~`case null` in `OTEntryPoint` - I don't think this will ever be null. I'll dive deeper and see if we can get the _underlying_ span to be null in this case.~ I'm worried this won't be universally defined for all OpenTracing implementations // update: turns out the context is actually null in this case. This is now handled in OTEntryPoint.


kubukoz:
@tpolecat have you seen this PR? I can resolve the conflicts and revive it but just wanted to make sure it's a desired change

mpilquist:
@kubukoz Any interest in refreshing this one?

kubukoz:
Not so much... I'm not sure about the exact state of OTel vs OpenTracing, and I don't know if I'll be using either anytime soon. I'll close for now.

PR #443, opened 2021-11-02
support natchez-noop on Scala.js

bpholt:
I think this was just an oversight. There might be others that can be supported too—I didn't look at them all.


tpolecat:
Thanks!

PR #448, opened 2021-11-12
`LiftTrace` concept

armanbilge:
~So this is not quite what we discussed on Discord, but I hope you'll like it better since you are suspicious of the two-parameter typeclasses. This does away with that by keeping the effect used for tracing _completely opaque_. This enables instances for both `Kleisli` and `IOLocal`.~

~IMO the ugliest part of it is that users will have to wrap their program in:~
```
trait Traceable[A] {
  def run[F[_]](implicit trace: Trace[F]): F[A]
}
```

~Everything subject to bikeshed of course. Let me know what you think!~


armanbilge:
Yeah, just ignore everything above, I bungled it 😅 obviously the problem with an opaque effect is that you don't have any other constraints available besides `Trace` ... which makes it pretty useless. So back to the original idea.

PR #452, opened 2021-11-17
Iterating on X-Ray support

bpholt:
This addresses some PR comments in #427 that we decided to hold up for a followup PR.

### DIFF
--- a/modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
+++ b/modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
@@ -226,7 +226,7 @@ private[xray] object XRaySpan {
       name: String,
       kernel: Kernel,
       entry: XRayEntryPoint[F],
-      useEnvironmentFallback: Boolean = true,
+      useEnvironmentFallback: Boolean,
   ): F[Option[XRaySpan[F]]] =
     OptionT.fromOption[F](kernel.toHeaders.get(Header))
       .subflatMap(parseHeader)
@@ -244,9 +244,10 @@ private[xray] object XRaySpan {
   def fromKernelOrElseRoot[F[_] : Concurrent : Clock : Random : XRayEnvironment](
       name: String,
       kernel: Kernel,
-      entry: XRayEntryPoint[F]
+      entry: XRayEntryPoint[F],
+      useEnvironmentFallback: Boolean,
   ): F[XRaySpan[F]] =
-    OptionT(fromKernel(name, kernel, entry))
+    OptionT(fromKernel(name, kernel, entry, useEnvironmentFallback))
       .getOrElseF(root(name, entry))
 
   def root[F[_]: Concurrent: Clock: Random](
--- a/modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
+++ b/modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
@@ -5,6 +5,7 @@
 package natchez.xray
 
 import cats._
+import cats.data._
 import cats.effect._
 import cats.syntax.all._
 import cats.effect.Resource.ExitCase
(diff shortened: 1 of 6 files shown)


armanbilge (approved):
LGTM :)

modules/xray/shared/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -221,27 +222,32 @@ private[xray] object XRaySpan {
...
+      .semiflatMap(fromHeader(name, _, entry))
+      .orElse {
+        OptionT.whenF(useEnvironmentFallback) {
+          XRayEnvironment[F]
+            .kernelFromEnvironment
+            .flatMap(XRaySpan.fromKernel(name, _, entry, useEnvironmentFallback = false))
+        }
+          .flattenOption
+      }
+      .value
```
christiankjaer, line 242: I think a better way is to just read the environment variables and construct the kernel before `fromKernel` is called, so that people can use whatever method they prefer for getting the trace context.

armanbilge: Sorry, I couldn't quite follow. Do you mean, we should only read from the environment variables once, during setup?

christiankjaer: Not exactly. When you use X-Ray you always know whether you should use some headers or the environment variables. I don't think the concept of a fallback will ever be necessary. The construction of the kernel is application specific and you probably never mix environment variables and HTTP headers for instance.

armanbilge: Oh, I see, thanks for explaining. So basically, expose this as a config option when creating the `EntryPoint`? Or do you mean, the `EntryPoint` itself should never grab the kernel from the environment, and it should be up to the user to do this manually?

christiankjaer: In the entry point you always pass the kernel when you want to create a new root span from an existing trace. How this kernel is constructed is [up to you.](https://tpolecat.github.io/natchez/reference/entrypoints.html#http4s-server-example)

armanbilge: Yeah, it makes sense :) the reason I liked this approach is:
1. Grabbing the kernel from the environment is bound to be a common default.
2. Furthermore, doing so let's you avoid writing code specific to XRay. But for sure, implicitly continuing a span from the environment is a bit opaque. And maybe this is a fake abstraction anyway!

Personally, I think this is worth having because of (1), but maybe just needs to be made more configurable.

bpholt: FWIW, a user can disable the environment fallback by passing `useEnvironmentFallback = false`. They could also explicitly get the kernel from the environment by using `XRayEnvironment[F].kernelFromEnvironment` and flatMapping into `fromKernel`.

> If I am using AWS lambda I will always be using environment variables. If I am writing a http4s service I will always be using HTTP headers.

I'm not sure this is right: e.g. if there is an Feral http4s API Gateway Lambda, then it should use any incoming tracing headers if they're present in the request, but if not, I think Lambda will still have a Trace ID in the environment that should be used instead?

armanbilge: Apologies if I missed something, but I think the only hole in the current config is if someone wants to raise an error (or create a root) if the provided kernel is lacking _even if_ the environment can in fact provide a Trace ID as a fallback. e.g., "we error/reject any http request without tracing headers".

> if there is an Feral http4s API Gateway Lambda, then it should use any incoming tracing headers if they're present in the request, but if not, I think Lambda will still have a Trace ID in the environment that should be used instead?

Agreed 👍

bpholt: > Apologies if I missed something, but I think the only hole in the current config is if someone wants to raise an error (or create a root) if the provided kernel is lacking even if the environment can in fact provide a Trace ID as a fallback

They can do this if the set the `useEnvironmentFallback` param to false, right?

armanbilge: I think this option is on a package-private method and furthermore not available in the natchez `Entrypoint` abstraction. So the only place a user can adjust this config is when they create the XRay entrypoint.

bpholt: Oh, right. Ok, so we need to propagate that option outward.

bpholt: I updated the PR with what I think were the suggested changes. Please let me know what you think!

armanbilge: 👍 yes, exactly what I had in mind, thanks :) @christiankjaer what do you think?

tpolecat:
This looks good, can you rebase or otherwise resolve?

armanbilge:
I'll do that!

tpolecat:
superseded by #468

Issue #514, opened 2022-02-13
Tracing resources


rossabaker:
Given this `Client`:

```scala
  type Client[F[_]] = Request => Resource[F, Response]
```

We would like to able to span the entire resource, with child spans for acquisition and release:

```
ambient
+- client
  +- acquire
  +- use
  +- release
```

Inspired by natchez-http4s and #19, we might try:

```scala
    def tracedClient[F[_]: MonadCancelThrow: Trace](client: Client[F]): Client[F] = { req =>
      Resource(Trace[F].span("client") {
        Trace[F].span("acquire")(client(req).allocated).map { case (resp, release) =>
          (resp, Trace[F].span("release")(release))
        }
      })
    }
```

But the span tree looks like this:

```
ambient
+- client
| +- acquire
+- use
+- release
```

We could introduce a more powerful `TraceResource` type class:

```scala
trait TraceResource extends Trace[F] {
  def spanResource(name: String): Resource[F, Unit]
}
```

Tracing a client becomes:

```scala
    def tracedClient[F[_]: MonadCancelThrow: TraceResource](client: Client[F]): Client[F] = { req =>
      TraceResource[F].spanResource("client") >>
      Resource(
        Trace[F].span("acquire")(client(req).allocated).map { case (resp, release) =>
          (resp, Trace[F].span("release")(release))
        }
      )
    }
```

Executable proof of concept:

```console
$ scala-cli test -S 2.13 https://gist.github.com/rossabaker/8872792b06bd84e5be8fae3c9caf8731
```

We could avoid the second type class by adding the method to `Trace`.  This would be a breaking change.  In addition to the new abstract method, we'd have to `mapK` the `Resource` on all the transformed instances, which would require a `MonadCancel` where there are currently weaker constraints.  All the base instances except `noop` already require `MonadCancel`, so maybe that's okay.

/cc @zmccoy

rossabaker:
I tried adding `spanResource` to `Trace`, and the `KleisliTrace` has sprained my brain. 🤕

rossabaker:
You can put yourself in a dark place with this if you use allocated and don't release in reverse order.  It's totally fine to create a child span from a finished span, but we lose access to the root span.  I'm worried how fs2 rescoping might play here:

```scala
  test("weird and misnested") {
    val tracer = new MockTracer
    for {
      root <- IO(MockSpan[IO](tracer, tracer.buildSpan("root").start()))
[...]

PR #526, opened 2022-02-28
Add Trace.spanR

rossabaker:
Adds a way to span a resource, from acquisition through release.

See #514.


modules/core/shared/src/main/scala/Trace.scala
```diff
@@ -140,7 +177,7 @@ object Trace {
 
   }
 
-  implicit def liftKleisli[F[_], E](implicit trace: Trace[F]): Trace[Kleisli[F, E, *]] =
+  implicit def liftKleisli[F[_]: MonadCancelThrow, E](implicit trace: Trace[F]): Trace[Kleisli[F, E, *]] =
```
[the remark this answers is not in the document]
armanbilge: I have a PR futzing with `Resource#mapK` constraints but I don't think it helps much here.
- https://github.com/typelevel/cats-effect/pull/2646

I think I encountered this working on Bayou, and I wondered if it was a good argument to have two `Trace` typeclasses after all.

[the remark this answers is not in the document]
rossabaker: I don't think that PR helps much here: you still need `MonadCancelThrow[F]`, and we're going `F ~> G` and `G ~> F` unless someone is more clever than me.  (Possible!  Nay, probable!)

The base instances, other than `noop`, all already require `MonadCancelThrow`, so I don't think this limits which effects can be traced.  But it could require a stronger constraint in a few places along the way.

modules/core/shared/src/main/scala/Trace.scala
```diff
@@ -205,15 +205,17 @@ object Trace {
...
     new Trace[StateT[F, S, *]] {
-
       def put(fields: (String, TraceValue)*): StateT[F, S, Unit] =
         StateT.liftF(trace.put(fields: _*))
 
       def kernel: StateT[F, S, Kernel] =
         StateT.liftF(trace.kernel)
 
+      def spanR[A](name: String)(r: Resource[StateT[F, S, *], A]): Resource[StateT[F, S, *], A] =
+        trace.spanR(name)(Resource.unit).mapK(StateT.liftK[F, S]) *> r
```
armanbilge, line 217: I haven't fully grasped the implications of these definitions, but is this much different than if we tried to implement `spanR(name: String): Resource[F, Unit]` instead?

rossabaker: I haven't tested this, but ...

- If the transformed instance is "stateful" (parents `acquire`, `use`, and `release`), so are these, and would be nicer expressed as `spanR(name: String): Resource[F, Unit]`.

- If the transformed instance is "local" (parents `acquire` and `release` but not `use`), we still span the resource, but parent none of the three phases.

- Local and stateful tracing are equivalent through the current `Trace` and the current abstraction works well until we introduce resources.  

- Resource tracing can be implemented either in a stateful subclass without local instances, or in the `Trace` constraint with an awkward signature and weird semantics for the transformers.

- If we resort to a subclass, libraries would be choosing the tradeoff on behalf of apps.

armanbilge: I see, thanks, it took me a while to grasp this. IIUC, for a "stateful" instance if you have
```scala
def spanR[A](name: String)(r: Resource[F, A]): Resource[F, A]
```
then you can get "for free"
```scala
def spanR(name: String): Resource[F, Unit] = spanR(name)(Resource.unit)
```
with the exact same semantics as if it had been defined in terms of `spanR(name: String): Resource[F, Unit]` to begin with.

However, for a stateless instance there _is_ a difference whether you use `spanR[A](name: String)(r: Resource[F, A])` or `spanR(name: String): Resource[F, Unit]`. So you want to use the former whenever possible, and only use the latter when necessary e.g. `Stream` and transformer instances.

So in summary, defining `Trace` as in this PR doesn't compromise the power of the stateful `IOLocal` instance, it will work exactly like it does for a Bayou-style `Trace`.

What it does compromise however is the library experience: since we want `spanR(name: String): Resource[F, Unit]` to be the second-choice, we don't directly offer it on the typeclass, even though it's easily derived in practice.

rossabaker: Yes:

|                       | acquire | release | use |
|-----------------------|---------|---------|-----|
| Stateful base         | ✅       | ✅       | ✅   |
| Stateful transformed  | ✅       | ✅       | ✅   |
| Stateless base        | 🤷       | 🤷       | 🚫   |
| Stateless transformed | 🚫       | 🚫       | 🚫   |

✅ - both `spanR`
🚫 - neither `spanR`
🤷 - `spanR(name: String)(r: Resource[F, A])`, but not `spanR(name: String)`

armanbilge: Coo, thanksl. This PR seems good then: it doesn't compromise the stateful `IOLocal` instance, while still being at least somewhat compatible with `Kleisli`. On the library side, you can trace any `Resource`-like thing either directly or with `trace.spanR(name)(Resource.unit)`. So everyone should be happy, right?

rossabaker: I'm still not convinced it's the right abstraction when we have a type class where a column mixes ✅ and 🚫.

I'll fix the other modules, unmark it as a draft, and we'll see if anyone else has any thoughts.

armanbilge: > I'm still not convinced it's the right abstraction when we have a type class where a column mixes ✅ and 🚫.

That would leave us with Stateless I I believe? Where you can span a resource-like thing but can't parent any of the acquire/use/release phases.

FWIW I'd like Stateless I even more. Then we can have `spanR(name: String): Resource[F, Unit]`!

rossabaker: Under Stateless I, we can never add fields to any phase of a resource span.  It is never ambient and can't be retrieved.

#527 is the most consistent and this has the broadest support.  I think that's the tradeoff we're stuck with.

armanbilge: That's assuming that we can't have "inconsistent" implementations of `Trace`, which I think is fine.

rossabaker: Correct.  This supports everything with one type class, with three distinct behaviors.  #527 supports one behavior, but drops the instances that can't support it once we upgrade to a resource constraint.

#527 still feels more right to me.  I'll undraft this and maybe we can find a mediator. 😆

rossabaker:
I am in favor of continuing with #527 over this one, particularly if https://github.com/typelevel/fs2/pull/2843 works out.

armanbilge:
Is natchez okay to take on an fs2 dependency?

rossabaker:
The repo already depends on it for xray.  Not sure it's smart for core to take it on, but an fs2 module wouldn't hurt.

(Half-baked idea: as @zmccoy is fond of saying, tracing is most important around network calls.  Having a natchez-fs2 that traced some of its io functions would be cool.)

rossabaker:
It turns out a `spanS` is a lot easier than `spanR` for the Kleisli instance.  `translate` gives us a hook to thread the state (here, just a String) through to the outputs:

```scala
    def spanS[F[_], A](name: String)(s: Stream[Kleisli[F, String, *], A]): Stream[Kleisli[F, String, *], A] =
      s.translate(
        new (Kleisli[F, String, *] ~> Kleisli[F, String, *]) {
          def apply[B](k: Kleisli[F, String, B]) = k.local(_ => name)
        }
      )

    val s = Stream.eval(Kleisli.ask[IO, String])
    (s ++ spanS("child")(s) ++ s)
      .compile
      .toList
      .run("root")
      .flatMap(IO.println)
```

prints `List("root", "child", "root")`.

rossabaker:
Tracing can be viewed as an exchange `F ~> F` of an untraced effect `F` with a traced effect `F`.  This is witnessed by the polymorphic function that results from partially applying `Trace.span`.

If we redefine `spanR` as something that returns a resource of `F ~> F`, we decouple how we propagate the span from the structure of transformed effects.  Thus, `spanR` supports a `Trace[Resource[F, *], *]`, `Trace[Stream[F, *], *]`, and any other "dynamically scoped" monad, all without introducing a second type class.

Expectation is that:

`spanR(name).surround(fa) <-> span(name)(fa)`

Caveats:
* I haven't tested any of these.
* The transformer instances still all require MonadCancelThrow.
* `Trace[Resource[F, *]]` is still only ambient for acquisition and release, not use.  That's the price of stateless instances.
* Any `Stream` interruption hacks would have to be universal, because we're not defining a specific `spanS`.

rossabaker:
The API for tracing a stream (example from https://github.com/armanbilge/bayou/issues/1) ends up looking like

```scala
  def stream(implicit trace: Trace[IO]): Stream[IO, Unit] = {
    Trace[Stream[IO, *]].span("new_root")(
      Stream(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
        .covary[IO]
        .chunkN(3)
        .flatMap { chunk =>
          Trace[Stream[IO, *]].span("child")(
            Stream.eval(trace.put("vals" -> chunk.toList.mkString(","))) >>
            Stream.sleep[IO](1.second)
          )
        }
    )
  }
```

To weave in and out of Stream, we have to accept the simpler `Trace[IO]` and then summon a `Trace[Stream[IO, *]]`.  We could just reference `trace` if we made `spanS` a derived method on `Trace`, but then we'd need a `MonadCancelThrow` member, and fs2 would be forced into natchez-core.

ryanmiville:
Just figured I'd share that we've been using a custom `Trace` at work for the last few months that is essentially a combination of Bayou and the `F ~> F` implementation here and it's been working well for us.

I've open sourced it [here](https://github.com/ryanmiville/natchex). I don't intend to promote _yet another tracing library_. I'd much rather these concepts make it into Natchez.

mpilquist:
Any thoughts on this design versus #527 now that some time has passed? @armanbilge @rossabaker

rossabaker:
I moved onto otel4s and don't remember all the nuances, but I think the basic tradeoff was that this behaves similarly between stateless and stateful instances, and #527 was able to make the resource span the parent of `use` at the expense of stateless instances (and an extra type class.)

Thanks for picking it up.

PR #648, opened 2022-11-17
Implement mapK syntax for EntryPoint and Span

alexcardell:


modules/core/shared/src/main/scala/natchez/syntax/package.scala
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2019-2020 by Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package natchez
+
+package object syntax
```
mpilquist, line 7: Instead of a syntax package, can we add these methods directly to the traits?

PR #667, opened 2022-11-28
add TraceableValue typeclass and base implementations

bpholt:
Introduces a lawless typeclass that is responsible for converting between the domain types used by a traced application and the `TraceValue` type that is used to track attribute values that can be attached to spans.

Using a typeclass approach instead of relying on implicit conversions has a couple advantages. My original motivation was to allow for a form of semi-automatic instrumentation using cats-tagless's [`Weave` effect](https://github.com/typelevel/cats-tagless/blob/6419316c90d6f94ea053891aff8e83bbfe99ac04/core/src/main/scala/cats/tagless/aop/Aspect.scala#L53-L70), which requires method parameter types all be constrained by a user-specified typeclass. This requirement means that we can generically convert a value of type `A: TraceableValue` to a `TraceValue` in positions where the compiler cannot prove that the existing implicit conversions would have worked.

I recently published [Dwolla/natchez-tagless](https://github.com/Dwolla/natchez-tagless) to facilitate this kind of semi-automatic instrumentation. (If this PR is merged, I'd modify that library to remove its `ToTraceValue` typeclass and use the `TraceableValue` I'm proposing here.) 

Another benefit comes with the use of refinement / opaque types, where the typeclass instance can format or redact a value before its inclusion as a span attribute.

### DIFF
--- a/modules/core/shared/src/main/scala/TraceValue.scala
+++ b/modules/core/shared/src/main/scala/TraceValue.scala
@@ -14,11 +14,47 @@ object TraceValue {
   case class BooleanValue(value: Boolean) extends TraceValue
   case class NumberValue(value: Number) extends TraceValue
 
-  implicit def stringToTraceValue(value: String): TraceValue = StringValue(value)
-  implicit def boolToTraceValue(value: Boolean): TraceValue = BooleanValue(value)
-  implicit def intToTraceValue(value: Int): TraceValue = NumberValue(value)
-  implicit def longToTraceValue(value: Long): TraceValue = NumberValue(value)
-  implicit def floatToTraceValue(value: Float): TraceValue = NumberValue(value)
-  implicit def doubleToTraceValue(value: Double): TraceValue = NumberValue(value)
+  implicit def viaTraceableValue[A: TraceableValue](a: A): TraceValue =
+    TraceableValue[A].toTraceValue(a)
 
+  @deprecated("use toTraceValue", "0.3.0")
+  def stringToTraceValue(value: String): TraceValue = StringValue(value)
+  @deprecated("use toTraceValue", "0.3.0")
+  def boolToTraceValue(value: Boolean): TraceValue = BooleanValue(value)
+  @deprecated("use toTraceValue", "0.3.0")
+  def intToTraceValue(value: Int): TraceValue = NumberValue(value)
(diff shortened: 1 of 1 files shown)


bpholt:
I explored trying to use `A: TraceableValue` in most places where `TraceValue` is used in Natchez today, but ultimately concluded that it wasn't worth it. For example, the [`put`](https://github.com/tpolecat/natchez/blob/9d74b8feb76c15139370c0317848cf0dca05756a/modules/core/shared/src/main/scala/Span.scala#L19) and [`log`](https://github.com/tpolecat/natchez/blob/9d74b8feb76c15139370c0317848cf0dca05756a/modules/core/shared/src/main/scala/Span.scala#L22) methods on `Span` both take varargs of `(String, TraceValue)` tuples. Replacing that with `(String, A)` works in the simple case, but we'd have to stop supporting varargs in this way to propagate that change everywhere. e.g. the [`put` in `DDSpan.attachError`](https://github.com/tpolecat/natchez/blob/9d74b8feb76c15139370c0317848cf0dca05756a/modules/datadog/src/main/scala/DDSpan.scala#L90-L95) could still work if we created several `put` methods with increasing arities (or introduced some shapeless-style shenanigans, like using `ProductArgs` to convert varargs to an `HList`). That gets trickier when the vararg is used to add an arbitrary list of attributes to the span. For the purposes mentioned above, I concluded it's not worth adding that extra complexity to clean up the `TraceValue`/`TraceableValue` distinction.

modules/core/shared/src/main/scala/TraceValue.scala
```diff
@@ -14,11 +14,47 @@ object TraceValue {
...
+  * parameter `A` to Natchez's `TraceValue`.
+  *
+  * You may want to use this to customize the formatting of a value
+  * before attaching it to a span, or to support adding tracing as a
+  * cross-cutting concern using aspect-oriented programming from a
+  * library such as cats-tagless.
+  *
+  * @tparam A The type to be converted to `TraceValue`
+  */
+trait TraceableValue[A] { outer =>
```
bpholt, line 44: Definitely open to bike-shedding over the names here. Right now I think `TraceableValue` is slightly better than `ToTraceValue`, which is what I used in our library. 🤷‍♂️

Issue #673, opened 2022-12-08
Log backend ignores trace ids from Kernel with non-UUID trace ids.


mvalle:
Given a kernel like this 
```
val kernel = Kernel(Map(
  "X-Natchez-Trace-Id" -> "non-uuid-trace-id", 
  "X-Natchez-Parent-Span-Id" -> "non-uuid-span-id"
))
```
The `natchez.log.Log.entryPoint.continueOrElseRoot`  function will drop the trace id, because it's expecting UUIDs. This makes it impossible to use the Log backend with other backends that produce non-UUID trace ids.

PR #677, opened 2022-12-09
#673 Allow non-uuid trace ids from kernel

mvalle:
Replace the UUIDs in LogSpan to allow a kernel like this to be correctly picked up.
```
val kernel = Kernel(Map(
  "X-Natchez-Trace-Id" -> "non-uuid-trace-id", 
  "X-Natchez-Parent-Span-Id" -> "non-uuid-span-id"
))
```


mpilquist:
Closed in favor of #685

PR #679, opened 2022-12-14
Fix `spanR` for `IOLocal`-based `Trace`

armanbilge:


modules/core/shared/src/main/scala/Trace.scala
```diff
@@ -79,9 +79,12 @@ object Trace {
...
           } yield new (IO ~> IO) {
             def apply[A](fa: IO[A]): IO[A] =
-              local
-                .set(child)
-                .bracket(_ => fa.onError(child.attachError(_)))(_ => local.set(parent))
+              local.get.flatMap { old =>
+                local
+                  .set(child)
+                  .bracket(_ => fa.onError(child.attachError(_)))(_ => local.set(old))
+              }
```
armanbilge, line 86: The `parent` span for this `Resource` is not necessarily the span that was installed on the `IO` at point of transformation.

mpilquist:
Thanks! A test for this would be welcome but shouldn't prevent merge. There's a bunch of infrastructure in `InMemorySuite` that I moved over from natchez-http4s.

PR #688, opened 2022-12-21
Support SpanKind and linked traces as Span options

bpholt:
OpenTelemetry has the concepts of span kinds and linked traces, both of which I'd like to be able to specify for Natchez traces. This is a first draft of adding them to the Natchez API via the `Span.Options` interface, along with expanding where span options can be set.

Right now only the core, OpenTelemetry, Jaeger, X-Ray, log, and no-op modules are updated to support the new attributes. I wanted to get feedback on the API before spending time figuring out how this maps into the other backends.

### DIFF
--- a/modules/core/shared/src/main/scala/EntryPoint.scala
+++ b/modules/core/shared/src/main/scala/EntryPoint.scala
@@ -14,20 +14,32 @@ import cats.effect.Resource
 trait EntryPoint[F[_]] {
 
   /** Resource that creates a new root span in a new trace. */
-  def root(name: String): Resource[F, Span[F]]
+  final def root(name: String): Resource[F, Span[F]] = root(name, Span.Options.Defaults)
+
+  def root(name: String, options: Span.Options): Resource[F, Span[F]]
 
   /** Resource that creates a new span as the child of the span specified by the given kernel,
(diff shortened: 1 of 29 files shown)


mpilquist:
Approach looks good to me!

bpholt:
The `Continuous Integration / Build and Test (ubuntu-latest, 3.2.1, temurin@8, rootJVM) (pull_request)` check seems to be hung on the "Cache sbt" step, so we may have to rerun it if it eventually times out. Other than that, I think this is now ready to review, so I'm going to move it out of Draft status.

modules/core/shared/src/main/scala/EntryPoint.scala
```diff
@@ -14,20 +14,32 @@ import cats.effect.Resource
 trait EntryPoint[F[_]] {
 
   /** Resource that creates a new root span in a new trace. */
-  def root(name: String): Resource[F, Span[F]]
+  final def root(name: String): Resource[F, Span[F]] = root(name, Span.Options.Defaults)
+
+  def root(name: String, options: Span.Options): Resource[F, Span[F]]
```
mpilquist, line 19: Would be good to add some docs about the use cases for `options` here. Nothing too extensive.

PR #709, opened 2023-01-26
split InMemory out into published testkit module

bpholt:
This will let other projects integrating with Natchez use the in-memory test infrastructure to test their own work.


bpholt:
I swear this passed in our CI before I created the PR 🙃

PR #712, opened 2023-01-30
Add mock to root aggregate and integrate latest API changes

daddykotex:


daddykotex:
Can I get a approval to run the workflow to make sure the PR is all right?

PR #776, opened 2023-04-15
Fix URL to site

ysthakur:
Looks like [tpolecat.org/natchez/](http://tpolecat.org/natchez/) doesn't work, so I changed it to [typelevel.org/natchez/](https://typelevel.org/natchez/) instead.


mpilquist:
Thanks, fixed in #782

PR #783, opened 2023-04-20
Add further Tag definitions

nefilim:
These are really following the OpenTelemetry Semantic Attribute specification, so not sure this is the best place for them but there are no tags defined under the OpenTelemetry module currently. 

Also - not sure how important it is to preserve backward compatibility for `def status_code` - could add an overloaded function with String parameter if deemed necessary. 

This PR will be followed with another PR in `natchez-http4s` to update invocation to `status_code`, once it's in the correct format (int64 vs string) - at least the AWS Xray collector will correctly export the value and it will be available in the trace data, currently it's just showing as 0.


mpilquist:
Looks good. I'd avoid the change to `status_code` signature if possible and if not, then add the overload for biocompat.

PR #805, opened 2023-05-14
Add 'spanId' to Trace interface

gstro:
Looks like `spanId` was omitted from the `Trace[F]` interface, probably by accident? According to [this section](https://typelevel.org/natchez/reference/spans.html#span-properties) of the docs, it should be accessible from the ambient spans using the `Trace` effect.

### DIFF
--- a/modules/core/shared/src/main/scala/Trace.scala
+++ b/modules/core/shared/src/main/scala/Trace.scala
@@ -104,7 +104,7 @@ object Trace {
         override def traceId: IO[Option[String]] =
           local.get.flatMap(_.traceId)
 
-        override def spanId: IO[Option[String]] =
+        override def spanId(implicit F: Applicative[IO]): IO[Option[String]] =
           local.get.flatMap(_.spanId)
 
         override def traceUri: IO[Option[URI]] =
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/Trace.scala
```diff
@@ -49,6 +49,11 @@ trait Trace[F[_]] {
     */
   def traceId: F[Option[String]]
 
+  /** A unique ID for this span, if available. This can be useful to include in error messages for
+    * example, so you can quickly find the associated trace.
+    */
+  def spanId: F[Option[String]]
```
armanbilge, line 55: If we want to preserve backwards compatibility we'll have to do something like this so that there is a default implementation. Otherwise any implementations outside of this repo will break.

```suggestion
  /** A unique ID for this span, if available. This can be useful to include in error messages for
    * example, so you can quickly find the associated trace.
    */
  def spanId(implicit F: Applicative[F]): F[Option[String]] = F.pure(None)
```

armanbilge: No, I don't think so. Anyone using `Tracer[F]` where there is no `Applicative[F]` would simply be unable to call this method.

gstro: Oh right, because the constraint is on the method 😅. Makes sense to me!

gstro:
@armanbilge Any chance to get this approved? Or someone else I should ping?

armanbilge:
> Any chance to get this approved?

Does this actually compile for you?

gstro:
> Does this actually compile for you?

Oof, no, sorry! I just assumed the recommended change would be enough. I'll work on getting this ready!

gstro:
Ok, good to go now!

gstro:
@armanbilge Sorry to ping again but this is ready for review!

armanbilge (approved):
Looks great, thank you! (sorry, I'm not an official maintainer here, but I am sure they will take a look soon :)

PR #818, opened 2023-06-06
Pass span options in RootsSpan

msosnicki:
Wanted to ask if there is any reason why the `Span.Options` are not passed from `RootsSpan`? Currently `RootsSpan` is only used for `IOLocal` based implementation, but if typelevel/natchez-http4s#34 is accepted, it will be needed for both `IOLocal` and `Kleisli` to propagate options from `RootsSpan` correctly.


kubukoz (approved):

armanbilge (approved):
I guess this was just an oversight?

PR #889, opened 2023-09-28
improve error message

EstebanMarin:
For some reason found the error message was not clear on the actionable and made explicit how to change the method to take away the compiler error.

### DIFF
--- a/modules/core/shared/src/main/scala/TraceValue.scala
+++ b/modules/core/shared/src/main/scala/TraceValue.scala
@@ -19,15 +19,15 @@ object TraceValue {
 
   @deprecated("use `viaTraceableValue(value)`", "0.3.0")
   def stringToTraceValue(value: String): TraceValue = StringValue(value)
-  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
+  @deprecated("use `viaTraceableValue(value)`", "0.3.0")
   def boolToTraceValue(value: Boolean): TraceValue = BooleanValue(value)
-  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
+  @deprecated("use `viaTraceableValue(value)`", "0.3.0")
   def intToTraceValue(value: Int): TraceValue = NumberValue(value)
-  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
+  @deprecated("use `viaTraceableValue(value)`", "0.3.0")
   def longToTraceValue(value: Long): TraceValue = NumberValue(value)
-  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
+  @deprecated("use `viaTraceableValue(value)`", "0.3.0")
   def floatToTraceValue(value: Float): TraceValue = NumberValue(value)
-  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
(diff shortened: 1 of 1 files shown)


modules/core/shared/src/main/scala/TraceValue.scala
```diff
@@ -17,17 +17,17 @@ object TraceValue {
   implicit def viaTraceableValue[A: TraceableValue](a: A): TraceValue =
     TraceableValue[A].toTraceValue(a)
 
-  @deprecated("use toTraceValue", "0.3.0")
+  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
   def stringToTraceValue(value: String): TraceValue = StringValue(value)
```
armanbilge, line 21: Thanks for the PR. `viaTraceableValue` looks like an implicit conversion so it doesn't seem like its intended to be called explicitly, so something still doesn't seem right. Would you be able to share the code you were having trouble with?

EstebanMarin: I cannot share the code unfortunately. Roughly is something like this
```
//> package natchez


/** A tracing effect, which always has a current span. */
trait Trace[F[_]] {

  /** Puts a sequence of fields into the current span. */
  def put(fields: (String, TraceValue)*): F[Unit]
somewhere in the code
Trace[F].put(", TraceValue.viaTraceableValue("string"))
```
Using something like `TraceableValue()` is not available. 



However, when encountering the error I assumed there was a new `.toTraceValue` method in the object, but there isn't. The only available `toTraceValue` comes from this [typeclass](https://github.com/estebanmarin/natchez/blob/35eb94d0513a1d10269d6ed073ba2d62a2861940/modules/core/shared/src/main/scala/TraceValue.scala#L44-L45) . After several rounds of compiler tetris I got this 
```
TraceValue.viaTraceableValue(paycenterAction)
```
To get my migration working from  < `0.3.0` to `0.3.3`.

armanbilge: Thanks for the explanation, much appreciated. I took a closer look and I agree this deprecation message is more suitable.

However, it looks like the `viaTraceableValue` implicit conversion should be applied automatically, and if it's not then that's a bug that we should fix. So if you could share an example that was working before and not working after the upgrade, that would be very helpful.

modules/core/shared/src/main/scala/TraceValue.scala
```diff
@@ -17,17 +17,17 @@ object TraceValue {
   implicit def viaTraceableValue[A: TraceableValue](a: A): TraceValue =
     TraceableValue[A].toTraceValue(a)
 
-  @deprecated("use toTraceValue", "0.3.0")
+  @deprecated("use .viaTraceableValue(TraceableValue)", "0.3.0")
```
armanbilge, line 20: I think this is a bit clearer, wdyt?
```suggestion
  @deprecated("use `viaTraceableValue(value)`", "0.3.0")
```

modules/core/shared/src/main/scala/TraceValue.scala
EstebanMarin: @armanbilge I agree. that implicit conversions should be applied automatically. I would love to work on that, let me set the code to replicate the issue. I will work on this.

armanbilge (approved):

PR #1017, opened 2024-05-24
Aws cloud log group name add in trace

ChristyGKrishnan:


modules/xray/src/main/scala/natchez/xray/XRaySpan.scala
```diff
@@ -115,7 +115,24 @@ private[xray] final case class XRaySpan[F[_]: Concurrent: Clock: Random](
         (goodKeys + ("malformed_keys" -> badKeys.keys
           .mkString(",")
           .asJson)) ++ fixedAnnotations
-
+      /*
+       * `aws_group_name` key in the annotations, when present, sets the AWS CloudWatch log group name.
+       * It is used to add log details to each trace.
+       */
+      val logGroupValue = allAnnotations.get("aws_group_name")
```
bpholt, line 122: Is this likely to be the same value for all spans in a given application? Or is it possible to vary on a per-span basis?

If it's going to be the same for all spans in an application, I wonder if it's better set as an option on [`XRayEntryPoint`](https://github.com/typelevel/natchez/blob/main/modules/xray/src/main/scala/natchez/xray/XRayEntryPoint.scala) (perhaps via [`XRayEnvironment`](https://github.com/typelevel/natchez/blob/main/modules/xray/src/main/scala/natchez/xray/XRayEnvironment.scala)) and then propagated here? Otherwise won't users have to set this every time they want to add a log to a span?

PR #1071, opened 2024-10-04
Add Cats and ScalaCheck instances

bpholt:
I'd like to use these in a forthcoming PR to natchez-http4s, and I thought they might be useful generally.


modules/core/shared/src/main/scala/Kernel.scala
```diff
@@ -21,4 +23,10 @@ final case class Kernel(toHeaders: Map[CIString, String]) {
 object Kernel {
   private[natchez] def fromJava(headers: util.Map[String, String]): Kernel =
     apply(headers.asScala.map { case (k, v) => CIString(k) -> v }.toMap)
+
+  implicit val kernelMonoid: Monoid[Kernel] =
+    Monoid.instance(Kernel(Map.empty), (a, b) => Kernel(a.toHeaders |+| b.toHeaders))
```
kubukoz, line 28: **issue:** `a.toHeaders` is a `Map[CIString, *]`, so its monoid will use the monoid of `String` to combine values under the same (case insensitive) key - it will concatenate them. I don't think these are safe semantics for this monoid.

```scala
scala-cli --toolkit typelevel:default
Welcome to Scala 3.5.0 (21.0.2, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> import org.typelevel.ci.*, cats.*, cats.implicits.*

scala> Map(CIString("aaa") -> "hello") |+| Map(CIString("AAA") -> "goodbye")
val res0: Map[org.typelevel.ci.CIString, String] = Map(AAA -> hellogoodbye)
```

I think the desired semantics would likely be to prefer RHS in case of conflict. So,

```suggestion
    Monoid.instance(Kernel(Map.empty), (a, b) => Kernel(a.toHeaders ++ b.toHeaders))
```

and at this point I would suggest a test for the conflicting case, just to prevent accidents.

bpholt: Good catch. I'll update accordingly.

bpholt: @kubukoz I updated the PR, if you want to take another look

kubukoz (approved):

PR #1108, opened 2024-12-16
Add an example of Trace instance for IO

mwz:
Hello, this PR adds a simple `Trace[IO]` example to the docs.
Would that be sufficient, or should I expand this to cover anything else?

### DIFF
--- a/modules/docs/src/main/paradox/reference/trace.md
+++ b/modules/docs/src/main/paradox/reference/trace.md
@@ -82,8 +82,6 @@ This is more general than the `Kleisli` instance above and allows you to instant
 
 ## Cats-Effect 3 IO Instance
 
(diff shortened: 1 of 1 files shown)


kubukoz:
I think the docs should use `Trace.ioTraceForEntryPoint`, it's more convenient.

It looks sufficient to me but I'd make that change^ :)

kubukoz:
sorry, I see I wasn't clear: I think the EntryPoint way should be the official recommendation, so it should go first. Is that OK with you?

mwz:
Sounds good—I've updated it 👍🏻

kubukoz (approved):

mwz:
@kubukoz, how do we get the website deployed?
Do I need to manually pr the generated static content against the `gh-pages` branch? It looks like it hasn't been updated in 3 years 🙀

kubukoz:
uh yeah I think we don't have any automation for the site. I'd rather we add a step that publishes to gh-pages as part of CI, using sbt-typelevel's github actions codegen; but regardless we should see if the build still works locally in the first place.

mwz:
ok no worries - I'm happy to look into that

PR #1119, opened 2025-01-17
Regenerate site

mwz:
I've regenerated the site manually using the latest version from main as it hasn't been updated in 4 years.

The only thing that doesn't work is the links to edit docs source files, which I can't quite figure out how to set up correctly in paradox, but they currently don't work anyway, so this is still an improvement over the existing version.

Would it make sense to move this site to `sbt-typelevel-site` and automate the deployments?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,49 @@
+*.class
+*.log
+
+# these are moved into the doc project by the build
+modules/docs/src/main/tut/changelog.md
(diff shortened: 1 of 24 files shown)


kubukoz:
>Would it make sense to move this site to sbt-typelevel-site and automate the deployments?

imo yes, I may not like the style but at least it'd be one thing less to maintain :)

PR #1186, opened 2025-06-23
Activate child spans just like root spans

mwisnicki:
`DDEntryPoint` activates created spans so `DDSpan` should too.

Activation binds span to current thread via ThreadLocal so it will become incorrect when IO/Fiber switches threads but at least it's an improvement and consistent with root spans.

Ideally there would be a Span option to disable activation entirely like this: https://github.com/typelevel/natchez/pull/1187

### DIFF
--- a/modules/datadog/src/main/scala/DDSpan.scala
+++ b/modules/datadog/src/main/scala/DDSpan.scala
@@ -74,6 +74,9 @@ final case class DDSpan[F[_]: Sync](
           case (span, ExitCase.Errored(e)) => Sync[F].delay(span.log(e.toString).finish())
           case (span, _)                   => Sync[F].delay(span.finish())
         }
+        .flatTap(span =>
+          Resource.make(Sync[F].delay(tracer.activateSpan(span)))(s => Sync[F].delay(s.close()))
+        )
         .map(DDSpan(tracer, _, uriPrefix, options))
     )
 


modules/datadog/src/main/scala/DDSpan.scala
```diff
@@ -74,6 +74,9 @@ final case class DDSpan[F[_]: Sync](
           case (span, ExitCase.Errored(e)) => Sync[F].delay(span.log(e.toString).finish())
           case (span, _)                   => Sync[F].delay(span.finish())
         }
+        .flatTap(span =>
+          Resource.make(Sync[F].delay(tracer.activateSpan(span)))(s => Sync[F].delay(s.close()))
+        )
```
mwisnicki, line 79: Copied from https://github.com/typelevel/natchez/blob/64c292de2ffa503c4459cc256cb121d87aef8e0e/modules/datadog/src/main/scala/DDEntryPoint.scala#L28-L30

kubukoz (approved):

mwisnicki:
As per https://github.com/typelevel/natchez/pull/1187#discussion_r2173608227 I now think that the right thing to do here is to never activate span since it is likely useless unless used as in https://github.com/typelevel/natchez/pull/1189

mwisnicki:
Obsoleted by https://github.com/typelevel/natchez/pull/1191

PR #1187, opened 2025-06-23
Add option to disable span activation

mwisnicki:
For now implemented only for Datadog.

Stacked on top of:

- [ ] https://github.com/typelevel/natchez/pull/1186

### DIFF
--- a/modules/core/shared/src/main/scala/Span.scala
+++ b/modules/core/shared/src/main/scala/Span.scala
@@ -175,7 +175,6 @@ object Span {
     def spanKind: Span.SpanKind
     def links: Chain[Kernel]
 
-    /** Whether to activate span in underlying tracing framework. This typically binds span to a ThreadLocal. */
     def shouldActivateSpan: Boolean
 
     def withParentKernel(kernel: Kernel): Options
@@ -186,6 +185,14 @@ object Span {
 
     def withLink(kernel: Kernel): Options
 
+    /** Whether to activate span in underlying tracing framework when starting the span.
+      *
+      * Note that span activation typically uses `ThreadLocal` and will be incorrect if your code shifts threads such as
+      * when using `IO`. Underlying tracing library might have some instrumentation to compensate for this, but it won't
+      * necessarily work correctly.
+      *
+      * This is the default behavior.
+      */
     def withShouldActivateSpan(shouldActivateSpan: Boolean): Options
   }
 


modules/core/shared/src/main/scala/Span.scala
```diff
@@ -175,13 +175,18 @@ object Span {
     def spanKind: Span.SpanKind
     def links: Chain[Kernel]
 
+    /** Whether to activate span in underlying tracing framework. This typically binds span to a ThreadLocal. */
+    def activateSpan: Boolean
```
kubukoz, line 179: **nitpick:** `shouldActivateSpan`, perhaps? The verb form gives me a little ick.

Or `spanActivationPolicy`, that'd be more consistent with the existing `spanCreationPolicy` param

modules/datadog/src/main/scala/DDEntryPoint.scala
```diff
@@ -52,7 +54,9 @@ final class DDEntryPoint[F[_]: Sync](tracer: ot.Tracer, uriPrefix: Option[URI])
     Resource
       .make(initSpan())(s => Sync[F].delay(s.finish()))
       .flatTap(span =>
-        Resource.make(Sync[F].delay(tracer.activateSpan(span)))(s => Sync[F].delay(s.close()))
+        Resource
+          .make(Sync[F].delay(tracer.activateSpan(span)))(s => Sync[F].delay(s.close()))
+          .whenA(options.activateSpan)
```
kubukoz, line 59: **question:** any other entrypoints/spans that should have this?

build.sbt
```diff
@@ -1,6 +1,6 @@
 import com.typesafe.tools.mima.core._
 
-ThisBuild / tlBaseVersion := "0.3"
+ThisBuild / tlBaseVersion := "0.4"
```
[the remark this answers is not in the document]
mwisnicki: I guess I would have to provide default noop implementations in the trait?
I.e. `def withActivateSpan(b: Boolean) = this`. Not sure if this makes sense.

I wish I knew how to test it without waiting for PR checks.

I've tried `sbt ++2.13.16 mimaReportBinaryIssues` as per https://typelevel.org/cats/contributing.html but it's not working:
```
[info] natchez-core: mimaPreviousArtifacts is empty, not analyzing binary compatibility.
```

modules/core/shared/src/main/scala/Span.scala
```diff
@@ -202,22 +207,25 @@ object Span {
         parentKernel: Option[Kernel],
         spanCreationPolicy: SpanCreationPolicy,
         spanKind: SpanKind,
-        links: Chain[Kernel]
+        links: Chain[Kernel],
+        activateSpan: Boolean
     ) extends Options {
       override def withParentKernel(kernel: Kernel): Options =
-        OptionsImpl(Some(kernel), spanCreationPolicy, spanKind, links)
+        OptionsImpl(Some(kernel), spanCreationPolicy, spanKind, links, activateSpan)
```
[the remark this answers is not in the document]
mwisnicki: changed to use copy

modules/core/shared/src/main/scala/Span.scala
```diff
@@ -202,22 +207,25 @@ object Span {
...
       override def withLink(kernel: Kernel): Options =
-        OptionsImpl(parentKernel, spanCreationPolicy, spanKind, links.append(kernel))
+        OptionsImpl(parentKernel, spanCreationPolicy, spanKind, links.append(kernel), activateSpan)
+      override def withActivateSpan(activateSpan: Boolean): Options =
+        OptionsImpl(parentKernel, spanCreationPolicy, spanKind, links, activateSpan)
     }
 
     val Defaults: Options =
-      OptionsImpl(None, SpanCreationPolicy.Default, SpanKind.Internal, Chain.empty)
+      OptionsImpl(None, SpanCreationPolicy.Default, SpanKind.Internal, Chain.empty, true)
```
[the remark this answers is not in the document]
mwisnicki: done

modules/datadog/src/main/scala/DDEntryPoint.scala
mwisnicki: Presumably other backends could use it but I'm not familiar with them and don't know if anyone uses them.

ThreadLocal based span activation on start is error prone so I'm not even sure if it's worth adding this option except it's the default behavior of datadog backend and I want ability to selectively disable it.

With opentelemetry but a better solution would be similar to https://github.com/typelevel/otel4s/pull/214

In general there really ought to be some kind of Java-interop method like [in otel4s](https://typelevel.org/otel4s/instrumentation/tracing-java-interop.html).

mwisnicki: Found a better solution to java-interop: https://github.com/typelevel/natchez/pull/1189

PR #1191, opened 2025-06-30
Do not activate span since thread-local propagation does not work correctly when using cats-effect

mwisnicki:
As mentioned in https://github.com/typelevel/natchez/pull/1186#issuecomment-3020004700 we should not be activating spans on thread local by default. This is incorrect when using `IO`.

The only safe way to do this is explicitly as in https://github.com/typelevel/natchez/pull/1189.

Not activating spans is also consistent with other backends like OpenTelemetry.


mwisnicki:
Any chance of getting this merged?

PR #1192, opened 2025-06-30
Add noop EntryPoint

mwisnicki:
I find it useful

### DIFF
--- a/modules/core/shared/src/main/scala/EntryPoint.scala
+++ b/modules/core/shared/src/main/scala/EntryPoint.scala
@@ -4,7 +4,7 @@
 
 package natchez
 
-import cats.~>
+import cats.{Applicative, ~>}
 import cats.effect.MonadCancel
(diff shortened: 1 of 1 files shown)


kubukoz (approved):
curious, what's the usecase? I figured most people would use `Trace.Implicits.noop` instead

mwisnicki:
> curious, what's the usecase? I figured most people would use `Trace.Implicits.noop` instead

I have runtime flag to enable tracing and use noopEntryPoint when tracing is disabled.

This simplifies main() slightly since I always use kleisli rather than having separate non-kleisli codepath for disabled tracing.

kubukoz:
Is there a particular reason you're using Kleisli? I thought IOLocal-based propagation would essentially replace that for CE3 users.

mwisnicki:
I'll switch to IOLocal eventually

PR #1335, opened 2026-07-03
Add PropagatingInMemory testkit backend

kubukoz:
## What

Adds `PropagatingInMemory`, a testkit `EntryPoint`/`Span` implementation that models trace **context propagation**, complementing the existing `InMemory`.

`InMemory` records a command log but gives every span an empty/inherited `kernel` and a `None` `spanId`, so it can't express *parentage* — you can't assert that a continued span is really a child of the span whose kernel was propagated to it. `PropagatingInMemory`:

- assigns each span a distinct id (via a `Concurrent` counter), exposed idiomatically through **`spanId`** and its **`kernel`** (under the `X-Natchez-Span-Id` header);
- resolves an explicit `parentKernel` (from `Span.Options.parentKernel`) over the enclosing span when creating a child — the semantic real backends implement;
- exposes the resolved parent as **`Span.parentId`**, and records each created span (`Recorded(name, id, parentId)`) queryable via `EntryPoint#spans`.

## Why

Testing that instrumentation propagates trace context across boundaries (e.g. a cache loader that continues a span from a kernel handed to it) currently isn't possible with the shipped testkit — every `InMemory` span hands out the same kernel and no id. This backend makes those assertions straightforward while leaving `InMemory` completely untouched.

## Notes / open questions

- New file, no changes to `InMemory` — all existing `core-tests` fixtures pass unchanged.
- `PropagatingInMemorySuite` demonstrates: distinct ids surfaced via `spanId`/`kernel`, child parented to enclosing span, explicit `parentKernel` winning over the enclosing span, and concurrent children each keeping their own parent with no id collision.
- Draft: only Scala 2.13/JVM was compiled locally; the source is written cross-friendly but JS/Native/2.12 are unverified. Naming (`PropagatingInMemory`, `Recorded`, `spans`) and whether this belongs as a sibling vs. an extension of `InMemory` are open to feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/core-tests/shared/src/test/scala/PropagatingInMemorySuite.scala
+++ b/modules/core-tests/shared/src/test/scala/PropagatingInMemorySuite.scala
@@ -1,0 +1,96 @@
+// Copyright (c) 2019-2020 by Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package natchez
+
+import cats.effect.IO
+import cats.syntax.all._
+import munit.CatsEffectSuite
+
+class PropagatingInMemorySuite extends CatsEffectSuite {
+
+  // Every span exposes a distinct id via spanId, and its kernel advertises that same id.
+  test("spanId is distinct per span and matches the id carried in the kernel") {
+    PropagatingInMemory.EntryPoint.create[IO].flatMap { ep =>
+      ep.root("root").use { root =>
+        root.span("child").use { child =>
+          for {
+            rootId <- root.spanId
+            childId <- child.spanId
+            childKern <- child.kernel
+          } yield {
+            assert(rootId.isDefined)
+            assertNotEquals(rootId, childId)
+            assertEquals(childKern.toHeaders.get(PropagatingInMemory.SpanIdHeader), childId)
+          }
+        }
+      }
+    }
+  }
+
+  // A child created inside a span is parented to that enclosing span.
+  test("child span's parentId is the enclosing span") {
+    PropagatingInMemory.EntryPoint.create[IO].flatMap { ep =>
+      ep.root("root").use { root =>
+        root.span("child").use { _ =>
+          ep.spans.map { spans =>
+            val root = spans.find(_.name == "root")
+            val child = spans.find(_.name == "child")
+            assertEquals(child.flatMap(_.parentId), root.map(_.id))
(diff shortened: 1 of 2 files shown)


bpholt:
This looks pretty useful! I was about to start working on something similar (I stumbled across a bug in some application code and wanted to use the lineages `InMemory.EntryPoint` to assert on span propagation, but as you know, that doesn't work) but luckily I noticed this PR first.

> whether this belongs as a sibling vs. an extension of InMemory are open to feedback

IMO this would be more useful as an extension or update to `InMemory.EntryPoint` instead of a separate backend. `InMemory`'s lineage concept works pretty well for what it does; it just doesn't support propagation via `continue`/`continueOrElseRoot`. But with this design, I think I'd need to use both `PropagatingInMemory` and `InMemory` in separate tests (or write some kind of bridging `EntryPoint` that delegates to both of them) and then have separate assertions. It'd be nice if it were all collapsed into one.

I suppose the downside is that existing tests with lineages might be broken? I wonder who's using that to test `continue`/`continueOrElseRoot` without wanting propagation to be checked; they might think they have coverage already and be missing issues?

modules/core-tests/shared/src/test/scala/PropagatingInMemorySuite.scala
bpholt, line 1: I noticed there are no tests that use `continue` or `continueOrElseRoot`. It'd probably be a good idea to test propagation via those paths too.

modules/testkit/shared/src/main/scala/PropagatingInMemory.scala
```diff
@@ -0,0 +1,133 @@
...
+
+    override def root(name: String, options: natchez.Span.Options): Resource[F, Span[F]] =
+      Resource.eval(newSpan(name, none, log, nextId, options))
+
+    override def continue(
+        name: String,
+        kernel: Kernel,
+        options: natchez.Span.Options
+    ): Resource[F, Span[F]] =
+      Resource.eval(newSpan(name, kernel.toHeaders.get(SpanIdHeader), log, nextId, options))
```
bpholt, line 113: There's a comment on `EntryPoint#continue` that says "If the required headers are not present in `kernel` an exception will be raised in `F`." I wonder if `SpanIdHeader` is not present in `kernel`, this should raise instead of silently falling back to effectively a new root span?

